PR #9, opened 2019-06-25
Removed Lieutenant requirement in newTools baseline

dupriezt:


dupriezt:
This requirement prevented loading the newtools baseline, because exception was signalled.
Exception: "MetacelloNameNotDefinedError: project group, or package named: 'Lieutenant' not found when used in requires: or includes: field or package: 'NewTools-Inspector-Extensions' for version: baseline of BaselineOfNewTools"

jecisc:
We need to add Commander2 as a dependency instead. I'll update it.

jecisc:
Done.

PR #58, opened 2020-02-27
Add dependency to the last version of Spec2

badetitou:
I suppose that the NewTools should depends on the last version of Spec2 during its dev.
(I also format with the default formatter the baseline method...)


StevenCostiou:
@badetitou does it load the latest incoming version of spec?
For example, the script I geenrally use has    
```Smalltalk
  onConflict: [ :e | e useIncoming ];
  onUpgrade: [ :e | e useIncoming ];
```


Also, maybe for such big method commit the change and the reformatting in two separate commits, because git considers that everything has changed  and it is hard to see what is the real change you made (e.g., everything is red & green).

StevenCostiou:
Seems to work.
I'm working a bit with a version loaded with this baseline to see if any problems.

PR #80, opened 2020-07-22
remove StInspectorContextIndexedTempNode

### LINKED ISSUE #78: StInspectorContextIndexedTempNode can be removed
StInspectorContextIndexedTempNode is not referenced. With the complexity of temp reading, StInspectorContextNamedTempNode is definitely the way to go.

StInspectorContextNamedTempNode should implement reading like this:

rawValue

	^ self hostObject tempNamed: tempName

This is much faster, as we do not need to create the list of all possibly accessible temps first.

MarcusDenker:
remove StInspectorContextIndexedTempNode, fixes #78


MarcusDenker:
reading the temp by name is much faster, e.g. for a simple case:


[|temp t2 t3 | temp := 1.  thisContext tempNamed:  #temp ] bench.
 "'156131.121 per second'"

[|temp t2 t3| temp := 1.  thisContext namedTempAt: 1 ] bench.
 "'86076.170 per second'" 

[|temp t2 t3| temp := 1.  thisContext temporaries at: #temp ] bench.
 "'40592.844 per second'"

PR #104, opened 2020-08-26
master

Ducasse:
New SystemReporter in Spec2


StevenCostiou:
Did you also added the packages to the baseline?
I will fix the conflicts (it's on the debugger), but not yet I have to push some work first and see how it goes.

PR #122, opened 2020-10-19
Fix visibility of workspace variables in debugIt expressions

dionisiydk:
Debugger interaction model should use context #lookupVar: instead of receiver to extract bindings.
Found variables are converted to foreignVariable to be compatible to any evaluation context.

The fix requires PR for Spec2 to have an affect: https://github.com/pharo-spec/Spec/pull/955

### DIFF
--- a/src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
+++ b/src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
@@ -22,7 +22,7 @@ StDebuggerContextInteractionModel >> behavior [
 { #category : #binding }
 StDebuggerContextInteractionModel >> bindingOf: aString [
 
-	^ (context lookupVar: aString) asForeignVariableFrom: context
+	^context lookupVar: aString
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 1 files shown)


dionisiydk:
I did not found existing tests for modified classes. If you will point me where to add them I will do

MarcusDenker (commented):
We should double-check that the call to lookupVar: in #hasBindingsOf: does not create bindingings in the requestor

dionisiydk:
> We should double-check that the call to lookupVar: in #hasBindingsOf: does not create bindingings in the requestor

These messages are always sent to an instance of Context. Therefore there is no any requestor there

dionisiydk:
I will push asDoItVariable renames after integration of https://github.com/pharo-project/pharo/pull/7710

src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
```diff
@@ -22,7 +22,7 @@ StDebuggerContextInteractionModel >> behavior [
 { #category : #binding }
 StDebuggerContextInteractionModel >> bindingOf: aString [
 
-	^ self doItReceiver bindingOf: aString
+	^ (context lookupVar: aString) asForeignVariableFrom: context
```
dionisiydk, line 25:
```suggestion
	^ (context lookupVar: aString) asDoItVariableFrom: context
```

dionisiydk:
And it's ready again for merge

MarcusDenker (approved):
Checked in latest Pharo9: this is missing, thus debugger is broken :-)

PR #127, opened 2020-11-02
125-When-a-StDebugger-is-closed-programmatically-it-is-leaked

### LINKED ISSUE #125: When a StDebugger is closed programmatically it is leaked
When a SPWindow is closed programmatically, the WindowClosed announcement is never generated so it is never removed from the SpApplication window collection.

To reproduce:

- Generate a MNU
- Create the method
- Do proceed.

tesonep:
Adding a conditional variable to keep track if the debugger has been proceed so it is not clear when closed the window.

Maybe it should be done in the debugger state model.

Fix #125

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -86,7 +86,8 @@ Class {
 		'extensionToolsNotebook',
 		'toolbarCommandGroup',
 		'debuggerActionModel',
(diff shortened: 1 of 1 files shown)


StevenCostiou:
@tesonep if it works, let's take it! It should be done in the state model I agree, could you open an additional "enhancement" issue just stating that so I'll remember to perfect this later?
Thanks!

PR #134, opened 2020-11-27
Replaced hard coded class references to StDebuggerInspector, StDebugg…

### LINKED ISSUE #110: No hard coded classes (in SpDebugger)
see https://github.com/pharo-project/pharo/pull/7311

No time for me this week to send a PR as laptop battery has swollen - I first need for a new one for repair
until I'm able to properly use the device again.

dupriezt:
…erContext and StDebuggerInspectorModel with overridable methods

Fixes #110

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -86,7 +86,8 @@ Class {
 		'extensionToolsNotebook',
 		'toolbarCommandGroup',
 		'debuggerActionModel',
-		'unsavedCodeChanges'
+		'unsavedCodeChanges',
+		'hasProceeded'
 	],
 	#classVars : [
 		'ActivateDebuggerExtensions',
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -86,8 +86,7 @@ Class {
 		'extensionToolsNotebook',
 		'toolbarCommandGroup',
 		'debuggerActionModel',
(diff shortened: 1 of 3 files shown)


src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -86,8 +86,7 @@ Class {
 		'extensionToolsNotebook',
 		'toolbarCommandGroup',
 		'debuggerActionModel',
-		'unsavedCodeChanges',
-		'hasProceeded'
+		'unsavedCodeChanges'
 	],
```
StevenCostiou, line 90: This variable is removed because it is not used?
I remember Pablo introduced it to fix a problem with spec when closing the window, but I don't remember if we redesigned it after and if that variable is just a remain.

StevenCostiou (approved):
For me it is ok, just need confirmation about the reason if the instvar remove

src/NewTools-Debugger/StDebugger.class.st
dupriezt: It got removed by accident.
Iceberg showed a ton of changes I didn't make.
I probably failed to uncheck this one when doing the commit.
I pushed another commit to re-add this variable.

StevenCostiou: ok, then for me we can merge it now. thanks!

StevenCostiou: (just waiting to see if the tests pass, but I think they never do on this repo I do not know why)

PR #142, opened 2020-12-14
7698-Fix-Browse-Debugger

MarcusDenker:
- #doBrowseClass and #doBrowseReceiverClass need to evaluate the block they get

- stackSelectionMethodContext needs to send #home to make sure it is at the method context

fixes https://github.com/pharo-project/pharo/issues/7698


StevenCostiou:
Marcus the CI is not building, so I cannot see (we should fix that some day). Can you just confirm that all debugger tests pass?

MarcusDenker:
Yes, tests are green

pavel-krivanek:
works :-)

StevenCostiou:
thanks :)

PR #146, opened 2020-12-15
AST inspector: The pane in the inspector showing the Dump is highlighting strangely

MarcusDenker:
The reason is that the SpCodePresenter is highlighting it as a method, but it is an expression.

#beForScripting" sets it up correctly.


StevenCostiou:
Could we add a test to check that?
I do not know if there is an accessor to check if the inspector is in scripting or method or I don't know mode.

MarcusDenker:
we could hand over the AST to the SpCodePresenter

- it would know if this is a method or not
- it would not need to re-parse

PR #151, opened 2020-12-17
Small improvement to #inspectionSourceCode

MarcusDenker:
with the latest Pharo9, the #methodClass was improved to do what we did here and we can just call it

NOTE: this needs a recent Pharo9 to work


MarcusDenker:
with the change in Pharo9 now active for some weeks, we can merge this.

NOTE: you need a current Pharo for this to work.

PR #184, opened 2021-03-23
Improve-Inspector

MarcusDenker:
Improve-Inspector:
	- IRMethod IR
	- MetaLink definiton
	- Simplify Variable users


Ducasse:
I would like to integrate it but the CI fails. Now I think that it never worked. :(

PR #212, opened 2021-04-14
OrderedDictionary should have the same presentation as a dictionary

guillep:
Add presentation for OrderedDictionary


guillep:
This was already done

PR #214, opened 2021-04-20
remove-some-unused-ivars

MarcusDenker:
I wanted to just remove some unused ivars, but it now wants to commit other changes, too

I have no idea why


MarcusDenker:
And I tried to remove class instance variables, these changes are not commited

PR #215, opened 2021-04-23
Fix-Debugger-Playground-Bindings

MarcusDenker:
This PR fixes https://github.com/pharo-project/pharo/issues/8803

The idea is that StDebuggerContextInteractionModel does not define any bindings by itself, therefore, #hasBindingOf: should return nil. #bindingOf: forwards the lookup to the context. It now makes sure to
not create a new binding in the requestor (should there be one) by using #lookupVar:declare: with false.

### DIFF
--- a/src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
+++ b/src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
@@ -22,7 +22,7 @@ StDebuggerContextInteractionModel >> behavior [
 { #category : #binding }
 StDebuggerContextInteractionModel >> bindingOf: aString [
 
-	^ (context lookupVar: aString) ifNotNil: [ :var | 
+	^ (context lookupVar: aString declare: false) ifNotNil: [ :var | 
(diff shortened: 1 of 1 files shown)


src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
```diff
@@ -51,8 +51,8 @@ StDebuggerContextInteractionModel >> doItReceiver [
 
 { #category : #testing }
 StDebuggerContextInteractionModel >> hasBindingOf: aString [
-
-	^ (context lookupVar: aString) notNil
+	"The debugger does not define new bindings"
+	^nil
```
dionisiydk, line 55: it seems it is an error. It should return false.

MarcusDenker: ups. yes

estebanlm:
new PR, please :)

estebanlm:
ok, I fixed it upstream (easier)

PR #216, opened 2021-04-23
175-Do-we-need-StInspectorProtoObjectNode

### LINKED ISSUE #175: Do we need StInspectorProtoObjectNode?
StInspectorProtoObjectNode has no comment.

My theory is that is was added to be able to debug subclasses of protoObject without having to send messages.

But we improved StInspectorSlotNode to do that by default: it reads slots via the slots API and thus we do not need
 StInspectorProtoObjectNode

MarcusDenker:
We do not need to read ivars from ProtoObject subclasses specially, as the default way is to read them reflectively via the slots meta objects

- remove StInspectorProtoObjectNode 
- use StNodeCollector in ProtoObject>>#inspectorNodes

Fixes #175


MarcusDenker:
AAARGH...

MarcusDenker:
will redo it

PR #218, opened 2021-04-23
cleanup-StInspectorProtoObjectNode 

### LINKED ISSUE #175: Do we need StInspectorProtoObjectNode?

MarcusDenker:
StInspectorProtoObjectNode was added to be able to debug subclasses of protoObject without having to send messages.

But we improved StInspectorSlotNode to do that by default: it reads slots via the slots API and thus we do not need
StInspectorProtoObjectNode

- remove StInspectorProtoObjectNode
- use StNodeCollector in ProtoObject>>#inspectorNodes

fixes #175


estebanlm:
I added the changes manually, thanks!

PR #223, opened 2021-04-29
Improve-Debugger-Playground-Bindings

MarcusDenker:
This is an improved version of https://github.com/pharo-spec/NewTools/pull/215

This makes syntax highlighting work correctly.

- the non-creating lookup is now done in hasBindingOf:
- #bindingOf: now first calls # hasBindingOf: to avoid creating new bindings in the requestor

### DIFF
--- a/src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
+++ b/src/NewTools-Debugger/StDebuggerContextInteractionModel.class.st
@@ -22,8 +22,11 @@ StDebuggerContextInteractionModel >> behavior [
 { #category : #binding }
 StDebuggerContextInteractionModel >> bindingOf: aString [
 
-	^ (context lookupVar: aString declare: false) ifNotNil: [ :var | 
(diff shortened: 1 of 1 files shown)


dionisiydk (approved):
Few thought @MarcusDenker.

I think that all these problems with auto variables in the workspace proves that the old trick with the #lookup: is really bad idea.  
Why anybody should remember that the lookup is not a safe operation with very surprizing side effect?

Alternatively the workspace can simply compile the doIt and produce "normal" undeclares for unknown variables (as usual). But with special compilation hook the workspace will substitute the global Undeclared registry with own local bindings.  
It will be a generally good improvement to avoid global Undeclared and replace it with "self environment undeclaredRegistry". Then the workspace will only need to provide specialized environment into the doIt compiler.

MarcusDenker:
Yes, nice idea... I will add it to the things to look at. We want to change (improve) Undeclared handling in general, we could do experiments in this direction then.

PR #224, opened 2021-04-30
[improvement] add Integer inspecting, based on GT version

hogoww:
Retain behavior from the GT version.
Let me know if that's the proper way to contribute :)


guillep:
THANKS

PR #236, opened 2021-05-26
StInspectorTempNode-rawValue-hardening

MarcusDenker:
This PR makes #rawValue more forgiving for failure cases:
- trap all Errors, as, e.g. when something goes wrong with looking up temps stored in the temp vector, we could have other error.
- use ##tempNamed: to force a re-lookup of the variable object used to read the data from the context.


Ducasse:
Did you use the image I zipped?
Do you still need it?

MarcusDenker:
Yes, I used to image to find what was wrong in your image, but I need to continue found a way to reproduce it to write a test.

I have saved the image locally, so you can remove it from the server.

Ducasse:
Tx.

PR #249, opened 2021-06-11
always select first available item in list 

estebanlm:


Ducasse:
excellent
Did you try if it helps for the flow? Because may be I was wrong?

estebanlm:
I don't know, maybe it helps even if it is not exact match, in any case it was like that in the old glamour so people (like you :P) are already used to that flow so it will not hurt.

Esteban

Ducasse:
Cool
for me when I see what I want on the top I just want to type enter!

estebanlm:
this thing produces some weird behavior in UI... I will hold the PR until I figure out what is the problem.

PR #253, opened 2021-06-20
Category cleanups and alignments

### LINKED ISSUE #252: Category cleanups and alignments

astares:
Fix  #252

### DIFF
--- a/src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
+++ b/src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
@@ -34,6 +34,12 @@ StDebuggerCommand class >> forSpecContext: anObject [
 		yourself
 ]
 
+{ #category : #testing }
+StDebuggerCommand class >> isAbstract [
+
+	^ self == StDebuggerCommand
+]
+
 { #category : #default }
 StDebuggerCommand class >> shortName [
 	^self defaultName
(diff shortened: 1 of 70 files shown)


src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
```diff
@@ -34,6 +34,12 @@ StDebuggerCommand class >> forSpecContext: anObject [
 		yourself
 ]
 
+{ #category : #testing }
+StDebuggerCommand class >> isAbstract [
+
+	^ self == StDebuggerCommand
+]
+
```
StevenCostiou, line 42: I find it odd to apply that everywhere: the debugger commands (and some in the same commit, like Sindarin) are not meant to be tested as "abstract".

Is that rule for abstract classes we have to apply everywhere if Pharo?
Is it not a bit heavy?
And what would be the usage for sindarin and stdebugger commands?
Except that which bothers me (but if that is normal, please dismiss my comment), the PR is simple and should be merged.

astares: Please read:
https://github.com/pharo-project/pharo/issues/9211

and https://twitter.com/seanjtaylor/status/1403939114915745793 when it comes to "consistency"

Also it is checked:
 - either by the user manually (which at the moment would wrongly return false)
 - or the browser 
 
image


Possibly in the future we have more lightweight options to tag a class as abstract (in fact we have a properties API for methods and classes - but they do not get serialized yet)

StevenCostiou (approved):

PR #270, opened 2021-08-27
Improving inspector on halt in sensible methods.

Ducasse:


Ducasse:
We should wait that the Pr in Pharo is integrated.

PR #274, opened 2021-09-02
 Fix-Debugger-CleanBlocks

MarcusDenker:
Fix [1 halt] value to work if cleanblocks are enabled.


estebanlm:
I am just going to trust you in this ;)

PR #277, opened 2021-09-06
Fix/276

guillep:
Use StObjectPrinter to correctly display errors when printing inspector values
Added tests for halt, error and DNU printing


guillep:

estebanlm:
mmm... there is a halt at: `StObjectWithPrintHalt>>#printOn:` which is obviously on purpose, but needs to be marked with `<haltOrBreakpointForTesting>` :)

PR #278, opened 2021-09-06
Fix/275

### LINKED ISSUE #275: inspector extra pane should not break when inspecting a strange object
For example I have a fileReference whose path is a bytestring and the inspector just breaks. 
The inspector should be ROBUST! Else better not have one. 

FileSystem >> absolutePath
	"Return the absolute of the receiver"
	^ self path isRelative
               ~~~~~~~~~~~~
                Raises a DNU because I have a bytestring and not path (ok this is my problem but the inspector should not break). 

		ifFalse: [ self path ]
		ifTrue: [ filesystem resolve: self path ]

S

guillep:
Fix #275

- Handling errors during inspector context creation
- Handling errors during inspector presentation creation
- Provide a debugInspectorPresentation when failure happens

### DIFF
--- a/src/NewTools-Inspector-Tests/StInspectorTest.class.st
+++ b/src/NewTools-Inspector-Tests/StInspectorTest.class.st
@@ -24,6 +24,84 @@ StInspectorTest >> tearDown [
 ]
 
 { #category : #tests }
+StInspectorTest >> testBuildPresentationWithDoesNotUndertandShouldReturnCodePresenter [
+
+
+	| collector pragma context |
+	collector := StInspectionCollector on: StInspectorMockObject new.
+	pragma := (StInspectorMockObject >> #methodWithDnuingPresentation) pragmas first.
+
+	context := collector contextFromPragma: pragma.
+	
+	"Smoke test -> creating an inspection presenter should not fail. Instead it should create an error presenter"
+	context newInspectionPresenter
+]
+
+{ #category : #tests }
+StInspectorTest >> testBuildPresentationWithFailureShouldReturnCodePresenter [
+
+
+	| collector pragma context |
+	collector := StInspectionCollector on: StInspectorMockObject new.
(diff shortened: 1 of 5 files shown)


guillep:

guillep:
Just for documentation purposes, here is a screenshot of what happens with this PR when a presentation has an error (either in the context method or the main presentation method).

image

guillep:
Should fix https://github.com/pharo-project/pharo/issues/9898

estebanlm:
same here, some methods needs to be marked with `<haltOrBreakpointForTesting>`

src/NewTools-Inspector-Tests/StInspectorTest.class.st
```diff
@@ -23,6 +23,84 @@ StInspectorTest >> tearDown [
...
+
+
+	| collector pragma context |
+	collector := StInspectionCollector on: StInspectorMockObject new.
+	pragma := (StInspectorMockObject >> #methodWithDnuingPresentation) pragmas first.
+
+	context := collector contextFromPragma: pragma.
+	
+	"Smoke test -> creating an inspection presenter should not fail. Instead it should create an error presenter"
+	context newInspectionPresenter
```
demarey, line 37: Missing assertion on the returned presenter type (same remark for next 2 tests)? 
```Smalltalk
self assert: context newInspectionPresenter isErrorPresenter
```

guillep: This test does not return an error presenter but a normal CodePresenter instance. So testing it is much more difficult. What I could test is that it is a code presenter that contains code that when evaluated will reproduce the exception. That is doable, worth it?

demarey: Checking the code looks a bit complex for an assertion. 
Maybe the test name confused me. If it was named testBuildPresentationWithDoesNotUndertandShouldNotRaiseError, it would be ok for me.

Issue #231, opened 2021-05-11
Allow to change the output port for an inspected object


hogoww:
For example when inspecting a dictionary, the old behavior (which I kinda want back) was to inspect the value instead of the selected association.
However I didn't find how to do that.

	^ aBuilder newTable 
		addColumn: (SpStringTableColumn 
			title: 'Key'
			evaluated: [ :each | StObjectPrinter asTruncatedTextFrom: each key ])
			beSortable;
		addColumn: (SpStringTableColumn 
			title: 'Value' 
			evaluated: [ :each | StObjectPrinter asTruncatedTextFrom: each value ])
			beSortable;
		items: self associations;
		yourself

PR #279, opened 2021-09-06
Fix selection of association dictionaries in the inspector

### LINKED ISSUE #234: Inspector does not allow to go from methodDictionary to compiledMethod
We have to go via association value instead of a nice pane.

guillep:
Fix #234
Works around #231


guillep:

PR #280, opened 2021-09-06
Fix dnu because of glamour removal

guillep:
Fix ugly inspection icons

image

This was due to a DNU caused by gt removal


Ducasse:
Ahhhh so it should be me that broke it.

PR #281, opened 2021-09-06
Improving the inspection of compiled methods

guillep:
- Hide the "Items" view that showed only a collection with itself
- raw presentation should really be raw, showing all literals including header and binding

Improve the inspector also to
 - do context selection using the correct selector (seems like it was a typo)
 - make context selector to be searched from the receiver object and not from the class


estebanlm:
... and same here. Some methods need to be marked as `<haltOrBreakpointForTesting>` :)

estebanlm:
this has now conflicts :(

PR #282, opened 2021-09-06
Improving responsiveness of spotter

tesonep:
- Removing remaining gt messages.
- The Query should be executed after a little delay after the keyboard input.
- When the query ends we need to remove always the spinner. 
- The menu options are not in the empty query.
- When the quey ends we need to flush pending results.


estebanlm:
"The menu options are not in the empty query."
why not?
the default behavior is to show menu when empty query, and I think is correct: 

image

tesonep:
Because I tried to do it more responsive when you open. I have removed the
help topics that were shown after and they were taking almost a second,and
I did the same for the menu.

We can put them back, I tried to copy the behaviour of osx finder that does
not show nothing at all.

Another alternative is to show nothing and some text to say how to use the
more advanced features, as some skins of gnome do does.

estebanlm:
now there are some errors there :(

PR #291, opened 2021-09-20
fix #290: Update StObjectPrinterTest>>testPrintDoesNotUnderstand to remove string literal

### LINKED ISSUE #290: Update StObjectPrinterTest>>testPrintDoesNotUnderstand to remove string literal
The fix for https://github.com/pharo-project/pharo/issues/9907 (https://github.com/pharo-project/pharo/pull/9982) has a failing test in `NewTools`.

Putting this here so I can reference in the PR. I intend to update the test so that it doesn't rely on a string literal, but rather constructs the message from the `MessageNotUnderstood` error.

gypsydave5:
Removes the string and replaces it with the `messageText` of the expected MessageNotUnderstood error.

### DIFF
--- a/src/NewTools-Inspector-Tests/StObjectPrinterTest.class.st
+++ b/src/NewTools-Inspector-Tests/StObjectPrinterTest.class.st
@@ -14,10 +14,13 @@ StObjectPrinterTest >> assertIsError: aCollection expectedMessage: aMessage [
 { #category : #tests }
 StObjectPrinterTest >> testPrintDoesNotUnderstand [
 
-	| printedValue |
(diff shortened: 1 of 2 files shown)


MarcusDenker:
Note: this PR will fail as the test requires a fix to the image (the sad story of cross-cutting changes)

gypsydave5:
> Note: this PR will fail as the test requires a fix to the image (the sad story of cross-cutting changes)

_Should_ still pass as the PR removes the hard coded string; test depends on the messageText of the MNU error, of whatever version. But I'm new here so I really don't know!

MarcusDenker (approved):

PR #292, opened 2021-09-20
Moves class side `#reset` methods into the `initialization` protocol.

gypsydave5:
Hi! 👋 - this PR moves all class side `#reset` methods in this repo over to the `initialization` protocol. This is to move https://github.com/pharo-project/pharo/issues/2607 forward.

Hope that makes sense!


gypsydave5:
🙇

PR #293, opened 2021-09-23
Pragma-inspector-source

MarcusDenker:
This PR adds a source pane to the inspector for Pragmas


src/NewTools-Inspector-Extensions/Pragma.extension.st
```diff
@@ -0,0 +1,11 @@
+Extension { #name : #Pragma }
+
+{ #category : #'*NewTools-Inspector-Extensions' }
+Pragma >> inspectionSourceCodeMethod [
+	<inspectorPresentationOrder: 30 title: 'Method'>
+
+	^ SpCodePresenter new 
+		beForBehavior: self methodClass;
```
estebanlm: `beForMethod: self method` will do the trick :)

PR #294, opened 2021-09-24
Adding debugger tests when saving quick methods

guillep:
Creating a quick method in the debugger should not break
 - if the stepIntoQuickMethod setting is on, we should stay in the quick method
 - if the stepIntoQuickMethod setting is off, we should go back to the caller context

Check also that the stack is not smashed (same as in #8897)

This PR requires changes in Pharo (DebugSession) coming

### DIFF
--- a/src/NewTools-Debugger-Tests/StDebuggerTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerTest.class.st
@@ -4,7 +4,9 @@ Class {
 	#instVars : [
 		'session',
 		'debugger',
-		'oldFastTDD'
+		'oldFastTDD',
+		'dbg',
+		'stackSizeBefore'
 	],
 	#category : #'NewTools-Debugger-Tests-Presenters'
 }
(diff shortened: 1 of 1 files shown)


guillep:
Finally after some discussions we decided:

If you edit a method, you should stay in that method regardless it being quick or not.
The fact that methods are quick is an implementation detail and should not be cared about when debugging

guillep:
Depends on https://github.com/pharo-project/pharo/pull/10040

estebanlm:
thanks, but I am sorry, I can't integrate this right now, I am waiting for Pharo10 to be stable (so the merge tests run) :)
be patient.

MarcusDenker (approved):
- Build is green for Pharo10
- This just adds a test

For me this looks ok to merge

estebanlm:
`testNoShadowedVariablesInMethods` is failing.

estebanlm:
but this is simple, I can merge and fix myself

PR #308, opened 2021-10-22
Fix debugger improvement method creation

StevenCostiou:
- protocol is not asked anymore when creating a method in TDD
- protocol can be modified at anytime using the icon near the method name in the code pane
- some refactoring
- testing and more testing

https://user-images.githubusercontent.com/26929529/138465914-b75fab96-0c82-49c2-b944-0dcc43f0fd62.mov


MarcusDenker:
Something odd happend here, I have no idea why this was closed

PR #309, opened 2021-10-23
9956 fixes A new method gets make when you try to use the debugger to make a new class

### LINKED ISSUE #9956: A new method gets make when you try to use the debugger to make a new class
**Bug description**
When I have a test and it uses `aNewClass new`   the new class and also a new method is made 

**To Reproduce**
1.  make a test which uses  setUp with `myClass := aNewClass new in it
2.  Press on the dot to run the tests. 
3.  the debugger pops up that the class and a new method cannot be understood
4.  press create 

**Expected behavior**
Only the class schould be made 

**Version information:**
 - OS: Windows 10
 - Version: 
 - Pharo Version  9

StevenCostiou:
Fixes https://github.com/pharo-project/pharo/issues/9956

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -456,22 +456,18 @@ StDebugger >> copyStackToClipboard [
 { #category : #actions }
 StDebugger >> createMissingClass [
 
-	| message exception variableNode |
-	self flag: #DBG_MISSINGTEST.
-	message := self interruptedContext tempAt: 1.
-	exception := self interruptedContext tempAt: 2.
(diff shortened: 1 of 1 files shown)


StevenCostiou:
I'm not sure if we can consider the tests are passing or not.
I'm not certain the one failing on Pharo-integration are related to my changes.

estebanlm:
it should not.
let's see.

estebanlm:
mmm... I need to remove Pharo9.0 (I thought it was already done). 
In anycase... this is working :)

ghost:
is this also backported to P9 because that one we use for the exercism track ?

PR #322, opened 2022-01-05
Fixes pharo-spec/pharo#10255 Slow inspector on large Strings

carolahp:
Limiting string display length in inspector when inspecting a String and a Collection of Strings.
Improves efficiency


carolahp:
Fix https://github.com/pharo-project/pharo/issues/10255
Fix https://github.com/pharo-project/pharo/issues/10256

PR #325, opened 2022-01-11
Implementing FontChooser in NewTools

tesonep:


tesonep:
The errors are because there is an older version of the class in Pharo. There is PR to remove them, we should integrate them together

PR #327, opened 2022-01-17
Improving Inspector when inspecting and debugging "large" objects

carolahp:
Backporting solution of issue https://github.com/pharo-project/pharo/issues/10255 to Pharo 9.0

- Shrinking string preview
- Limiting string display length in inspector when inspecting a String and a Collection of Strings.


tesonep:
This PR was already integrated in Pharo10 branch. If we backport it to Pharo9, we should backport the whole Newtools version

PR #329, opened 2022-01-17
Replacing line breaks (cr and lf) with spaces

carolahp:
Replacing line breaks (cr and lf) with spaces


estebanlm:
this is conflicting :)
what's the rationale of this change?

estebanlm:
I still do not understand. Why it is supposed to happen?
(trying to understand, not saying no :P)

carolahp:
No problem :) 
I think that's the expected behavior because the "twin" method ```StObjectPrinter class >> asTruncatedTextFrom:``` does this

image

PR #334, opened 2022-01-20
333-StPlaygroundInteractionModel-needs-correctFromtowith

### LINKED ISSUE #333: StPlaygroundInteractionModel needs #correctFrom:to:with:
If you write 

Obj

select it and "print it", you get a DNU

MarcusDenker:
This PR adds correctFrom:to:with:

fixes #333

### DIFF
--- a/src/NewTools-Playground/StPlaygroundInteractionModel.class.st
+++ b/src/NewTools-Playground/StPlaygroundInteractionModel.class.st
@@ -62,6 +62,11 @@ StPlaygroundInteractionModel >> bindings [
 	^ joinedBindings
 ]
 
+{ #category : #'interactive error protocol' }
+StPlaygroundInteractionModel >> correctFrom: anInteger to: anInteger2 with: aString [ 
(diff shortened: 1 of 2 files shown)


src/NewTools-Playground/StPlaygroundInteractionModel.class.st
```diff
@@ -62,6 +62,11 @@ StPlaygroundInteractionModel >> bindings [
 	^ joinedBindings
 ]
 
+{ #category : #'interactive error protocol' }
+StPlaygroundInteractionModel >> correctFrom: anInteger to: anInteger2 with: aString [
```
estebanlm: can't we have better names for those variables?

estebanlm (changes requested):

src/NewTools-Playground/StPlaygroundInteractionModel.class.st
MarcusDenker: Done!

PR #347, opened 2022-02-17
Spotter filtering does not split the search string accordingly to camel syntax 

carolahp:
This change makes the query results more precise by preventing spotter to split the search string accordingly to camel case.

For example:
Before: the query 'StoreDisk' returned 'DiskStore'
Now: the query 'StoreDisk' does not return 'DiskStore'

### DIFF
--- a/src/NewTools-Spotter-Processors-Tests/StImplementorsProcessorTest.class.st
+++ b/src/NewTools-Spotter-Processors-Tests/StImplementorsProcessorTest.class.st
@@ -27,6 +27,13 @@ StImplementorsProcessorTest >> testFindThisMethodAfterAddingText [
 ]
 
 { #category : #tests }
+StImplementorsProcessorTest >> testFindThisMethodRespectingSearchStringTokensOrder [
+
+	self runForText: 'testFindMethodThis'.
+	self denyResultsIncludes: thisContext method
+]
+
+{ #category : #tests }
 StImplementorsProcessorTest >> testPopularStringDoesReturnOnlyTenResults [
 
 	self runForText: 'test'.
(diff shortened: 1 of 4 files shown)


tesonep (changes requested):
I don't understand why is it best to have like this.
What are the use cases that are improved?

carolahp:
> What are the use cases that are improved?

Users normally use camelcase to search (because it's the convention for names in Pharo) while not knowing that when the query is written in camel case, it is split as if it was separated by spaces (eg: the query 'StoreDisk' is equivalent to 'Store Disk').

Because of this behavior, Spotter returns too many irrelevant results when users use camel case to search.
This implementation makes spotter results more precise when users use camel case, which is most of the time. If users want to search different words, they still can use spaces between words

MarcusDenker:
I got at PharoDays the comment, too, that spotter shows too many useless hits to the point of being useless (for that programmer)

jecisc:
I sometimes find that useful. 
Often I know globaly what I'm looking for but I don't know the exact name. Writing what I think it is with camel case help me when I don't put things in the right order. 

But maybe those kind of results should be sorted after what matches exactly what we enter.

demarey:
@carolahp @tesonep What do we do with this PR?
My opinion is that spotter does not work well. If you know the exact selector and type it in spotter, you have a lot of noise around (selectors or classes with pieces of the selector or more).
I would be in favor to do not split on camel case. In this case, this PR could be ok (but some commented code needs to be removed). The previous search behaviour is still usable by using spaces instead of camel case

Ducasse:
Ping!

Ducasse:
Can we take a decision on this issue?

MarcusDenker:
This was now merged in the Pharo10 branch.

PR #361, opened 2022-04-19
Fixing initial window size in debugger presenter

### LINKED ISSUE #360: Initial window size of debugger is hard coded
In the method below, the initial size is hard coded, which is a code smell. As a consequence, the former code of this method is unreachable.

Historically, this was done because the debugger used to open in a very small window even on large screens. However, this trick isn't needed anymore and can be removed.

```Smalltalk
StDebugger >> initialExtent [
        "This code was taken from the old debugger to get the same behavior. Fell free to refactor"
	self flag: 'Do it better'.
	"RealEstateAgent standardWindowExtent y < 400"true "a tiny screen" 
		ifTrue: [ ^ 1000@800 ].
	
	^ [ | w h |
		w := Display width.
		h := Display height.
		(w - (w // 3)) @ (h - (h // 5)) ]
		on: Error
		do: [ 850@650 ]
]
```

adri09070:
Fixes #360 
In the method StDebugger >> initialExtent, the initial size of the debugger window was hard coded and the former code of the method was unreachable.

Historically, this was hard coded because of a bug: a small window was opened for the debugger, even on large screens. However, this bug doesn't seem to be there anymore so this can be removed

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -605,10 +605,6 @@ StDebugger >> forceSessionUpdate [
 
 { #category : #api }
 StDebugger >> initialExtent [ 
-	"This code was taken from the old debugger to get the same behavior. Fell free to refactor"
-	self flag: 'Do it better'.
-	"RealEstateAgent standardWindowExtent y < 400"true "a tiny screen" 
-		ifTrue: [ ^ 1000@800 ].
 	
 	^ [ | w h |
 		w := Display width.


adri09070:
It doesn't look like the failures are linked to my changes

StevenCostiou (approved):
Tests are passing.
Requires user feedback.

StevenCostiou:
@MarcusDenker can you have a look?
I think it could be merged.

MarcusDenker (approved):

PR #368, opened 2022-05-05
Migration of events from the debugger UI to the debugger action model to automate its updates

### LINKED ISSUE #367: The debugger action model is not updated automatically, needs refactoring
When an instance of `StDebugger` is notified with an event, the debugger action model is not updated automatically. This is the case, when it is notified that the session's state has changed, or when a new method has been created. For instance:
 
```Smalltalk
updateStep
	self flag: #DBG_UPDATE_OF_MODEL_SHOULD_BE_AUTOMATED.
	debuggerActionModel updateTopContext.
	debuggerActionModel updateContextPredicate.	
	debuggerActionModel updateDebugSession.
		
	self updateStackFromSession: self session.
	self updateWindowTitle.
	self updateToolbar.
	self flag: #DBG_INSPECTOR_UPDATE_BUG.
	inspector getRawInspectorPresenterOrNil ifNotNil: [ :p | p update ]
```

This is a problem because the debugger action model is part of the debugger's logic, so it should be able to exist without a GUI.

Furthermore, the session is encapsulated in a session holder, through which the debugger accesses the session. However, it is not needed as the debugger action model can now access the session, so it could notify the GUI when the session has changed. As a result, all session events and system announcement events should be moved to the action model so that it can update automatically.

To finish with, each session event triggers the call of a specific method but all these methods call the same method: `updateStep`. For instance:

```Smalltalk
updateStepInto
	self updateStep
```

Methods like these are unused hooks so they can be removed in the process of the refactoring

adri09070:
Fixes #367

In `StDebugger`, the session was accessed through a session holder. So, it was subscribed to the session holder to update itself and the action model when the session state - or the session itself - had changed. However, it should not be the case, as the action model should be able to exist without a GUI. Furthermore, the action model has access to the session so it could play the role of a session holder...

As a result, the action model is now subscribed to the session and handles all events that were handled in `StDebugger` before and `StDebugger` is now subscribed to the action model, which notifies the GUI when it has to update.
This migration of events was also the occasion to remove unused update hooks related to these updates.

Some Sindarin commands used to unsubscribe the debugger from the session to perform several actions and update the debugger only once. This is now handled by an "update guard" system in `StDebuggerActionModel`, notably thanks to the `preventUpdatesDuring:` method.

The handling of the `MethodAdded` event from the SystemAnnouncer, has been migrated from `StDebuger` to `StDebuggerActionModel` for the same reasons.


StevenCostiou:
@MarcusDenker this one is sensible, it requires multiple reviews.
I will do one, can you have a look too?

MarcusDenker:
Now there is a conflict

StevenCostiou:
@adri09070 can you check?

adri09070:
It doesn't look like the failing test concerns me

estebanlm:
so, can we merge @MarcusDenker ?

adri09070:
There were conflicts so I fixed them but now, there are a lot of errors.

It involves Collections so I don't think this is my fault but maybe I'm wrong

adri09070:
A test doesn't pass because of an undeclared variable in `Microdown class`...

MarcusDenker:
for the Undeclared in Microdown, I just opened an issue tracker entry: https://github.com/pillar-markup/Microdown/issues/485

(this issue can proceed without needing to wait it to be fixed)

MarcusDenker (approved):

PR #374, opened 2022-05-25
Refactoring of the `StDebugger` layout from class side to instance side to make it dynamic

### LINKED ISSUE #363: Refactor the debugger so that it uses a dynamic layout with spec
In the debugger presenter (StDebugger), the layout is defined statically with class methods.

Spec 2 provides more simple ways to do it dynamically, on the instance side. This would make the debugger's GUI more extensible

As a result, all methods in the '#specs' category and in the '#layout' category should be moved from class side to instance side to use a dynamic layout.

adri09070:
Fixes #363 
Using the old Spec API, the `StDebugger` layout was defined statically, on class side.

Now, it is defined dynamically on instance side, thanks to the new Spec2 API.

In the future, it will be useful to dynamically update the debugger layout each time a debugger extension is action/desactivated


adri09070:
The CI fails due to a test class that was testing the mechanism to choose the spec selector to use for the `StDebugger` layout.

Now that the debugger layout is dynamic, this mechanism doesn't exist anymore, as it only relies on the `stackAndCodeContainer` layout. So, in my opinion, this test class should be removed.

StevenCostiou:
@adri09070 Tests should be updated.
I guess for now they could be removed since you removed the logic.
By default, make sure the extension column is deactivated.

StevenCostiou:
> The CI fails due to a test class that was testing the mechanism to choose the spec selector to use for the `StDebugger` layout.
> 
> Now that the debugger layout is dynamic, this mechanism doesn't exist anymore, as it only relies on the `stackAndCodeContainer` layout. So, in my opinion, this test class should be removed.

Ah, sorry, I wrote my previous message before reading your comment.
Yes I think you are right.

StevenCostiou:
> By default, make sure the extension column is deactivated.

By reviewing the changes, i also saw that this is already done :)

adri09070:
I deleted the test class and an usused class variable in StDebugger and now tests pass

MarcusDenker:
There is now a conflict

adri09070:
It was just a conflict between two different methods that can coexist.
It doesn't look like the failing test concerns me

MarcusDenker (approved):

PR #376, opened 2022-05-25
Adding `debuggerLayout` in TStDebuggerExtensionLayout` to be able to choose the layout the debugger extension should have

### LINKED ISSUE #375: TStDebuggerExtension does not allow to specify the layout that has to be used in the debugger for debugger extensions.
It should have a `debuggerLayout` so that each debugger extension is not bound to use its default layout

adri09070:
Fixes #375 
Any debugger extension was bound to have its default layout in the debugger.

I've made an improvement so that any debugger extension can specify the layout it will have in the debugger, which is the default layout by default, but could be a totally different layout


StevenCostiou:
The test that does not pass is a helper class for tests.
It should be fixed inside so that the test passes.

adri09070:
I've specified a layout for the dummy extension presenter (for tests) and tests pass

PR #382, opened 2022-06-24
[improvement] Debugger Commands shortcuts are now displayed in tooltip

hogoww:
When passing the mouse on a button for a debugger action, it wasn't displaying the shortcut.
Many people do not know it (Myself included for reasons).
This PR adds the shortcut inside the description of the debug action in the tooltip

### DIFF
--- a/src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
+++ b/src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
@@ -103,7 +103,13 @@ StDebuggerCommand >> initialize [
 		stream 
 			<< self class defaultName
 			<< ': '
-			<< self class defaultDescription ])
+			<< self class defaultDescription.
+			self class defaultShortcut ifNotNil: [ :shortCut |
+				stream << '('	
+				<< shortCut printString
+				<< ')'  
+				]
+			])
 ]
 
 { #category : #'as yet unclassified' }


src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
```diff
@@ -103,7 +103,13 @@ StDebuggerCommand >> initialize [
 		stream 
 			<< self class defaultName
 			<< ': '
-			<< self class defaultDescription ])
+			<< self class defaultDescription.
+			self class defaultShortcut ifNotNil: [ :shortCut |
```
StevenCostiou, line 107: Could we return a null shortcut object when calling default shortcut in order to  avoid nil checks?

StevenCostiou (changes requested):

src/NewTools-Debugger-Commands/StDebuggerCommand.class.st
StevenCostiou: Looks like the same work was done in parallel ;)
https://github.com/pharo-spec/NewTools/pull/384

hogoww: Cool :')
We all talked about it on discord a while back, so not surprising.

This could of course be done, but it requires changing a subclassResponsibility.
Could also implement it in every subclasses to return the nil shortcut, but i don't know much about it

MarcusDenker:
https://github.com/pharo-spec/NewTools/pull/384  has been merged, I close this as a duplicate

PR #393, opened 2022-07-08
Updating Sindarin to Pharo-11 branch

adri09070:
BaselineOfNewTools was importing the Pharo-10 branch of the https://github.com/pharo-spec/ScriptableDebugger repository.

I have reworked this repository on a Pharo-11 branch so the changes made on this branch are not synched with what's imported by NewTools.

Plus, the CI used to be red because of Sindarin tests but I have fixed them on the Pharo-11 branch so it looks like it's a good time to change the version of Sindarin that must be imported :)


StevenCostiou:
@estebanlm @MarcusDenker For me this is ok.
Pharo-11 branch is the new Sindarin integration branch for Pharo.
I would like your approval before merging because it will be the new branch loaded into Pharo by default.

MarcusDenker (approved):

Issue #91, opened 2020-08-10
Can't load NewTools-FileDialog


ackerleytng:
I'm getting a `#asTraitComposition was sent to nil`, where `Smalltalk compiler evaluate: self traitCompositionString` is nil.

To reproduce, I'm using a Pharo 9.0 image and after doing this

```
Metacello new
    baseline: 'NewTools';
    repository: 'github://pharo-spec/NewTools';
    load.
```

I manually loaded `NewTools-FileDialog` and got this issue.

I noticed that [`TSpDynamicPresenter`](https://github.com/pharo-spec/NewTools/blob/6843c1b6c8e13ccd9eea3103afa2306308db41b1/src/NewTools-FileDialog/FDPathPresenter.class.st#L23) is nowhere in this repository, could that be why?

StevenCostiou:
Hi @ackerleytng it is possible you need to also load https://github.com/pharo-spec/Spec

ackerleytng:
@StevenCostiou I tried loading Spec with the Metacello code on the Spec readme, but I'm still getting the same error. I tried editing the .class.st file directly, and I'm able to load FileDialog if I delete these two lines

```
	#traits : 'TSpDynamicPresenter',
	#classTraits : 'TSpDynamicPresenter classTrait',
```

Is there something else I can try?

ackerleytng:
Here's the full stack in the debugger, which I get when I load FileDialog.

```
UndefinedObject(Object)>>doesNotUnderstand: #asTraitComposition
UndefinedObject>>doesNotUnderstand: #asTraitComposition
MCClassDefinition>>traitCompositionCompiled
[ :builder | 
			builder
				superclass: superClass;
				name: name;
				layoutClass: (ObjectLayout layoutForType: type);
				slots: self instanceVariables;
				sharedVariables: self classVariables;
				sharedPools: self sharedPoolsString;
				classSlots: self classInstanceVariables;
				traitComposition: self traitCompositionCompiled;
				classTraitComposition: self classTraitCompositionCompiled;
				comment: comment stamp: commentStamp;
				category: category;
				environment: superClass environment ] in [ Smalltalk classInstaller
		make: [ :builder | 
			builder
				superclass: superClass;
				name: name;
				layoutClass: (ObjectLayout layoutForType: type);
				slots: self instanceVariables;
				sharedVariables: self classVariables;
				sharedPools: self sharedPoolsString;
				classSlots: self classInstanceVariables;
[...]

PR #396, opened 2022-07-18
FileBrowser loadable again: Moved to Spec2, P10+ and more fixes

JanBliznicenko:
Rework of FileBrowser/FileDialog.
There were several issues with it: Some previous refactorings were only half-done, it was based partially on old version of Spec2 and some parts even on Spec1 and needed multiple changes for Pharo 10-11 compatibility (originally compatible with Pharo 8).
Fixes also #91


JanBliznicenko:
Sorry for the force pushes, damn those line endings...
Btw. I actually use this version FileBrowser in the OpenPonk tool.

PR #402, opened 2022-07-27
Adding debugger extension visitor to display variables from extensions (as Chest)

adri09070:
This follows the PR that modified `StDebuggerContextInteractionModel` to allow to add bindings into the debugger itself: #378. This concept is used for example by Chest, which will be integrated soon into Pharo and allows to load any object from anywhere in Pharo into the debugger: https://github.com/adri09070/Chest

Now that we can add bindings into the debugger itself,  I made this PR to see these bindings in the debugger inspector and to be able to distinguish those that were added from those that come from the context.

To do that, I created an extensible debugger extension visitor system. So, if any other debugger extension add variables into the debugger, we will be able to distinguish those which debugger extension added which variable into the debugger .

Plus, `StDebuggerExtensionVisitor` provides hooks for every debugger extension and can be easily subclassed to perform other operations on debugger extensions.

The only doubt that I have is that I define the `visitChest:` method in `StDebuggerExtensionVisitor` and in `StDebuggerExtensionInspectorNodeBuilderVisitor`. As Chest is not in NewTools, I wonder if it would not be better to put this method as an extension method in Chest (let me know what you think but anyway, it allows you to see how it will work in Chest)

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -1245,14 +1245,19 @@ StDebugger >> updateCodeFromContext [
 { #category : #'presenter - code' }
 StDebugger >> updateCodeFromContext: aContext [
 
-	| selectionInterval |
+	| selectionInterval formerActionModel |
 	aContext ifNil: [ ^ self clearCode ].
 
 	self recordUnsavedCodeChanges.
 	selectionInterval := self selectedCodeRangeForContext: aContext.
 	aContext sourceCode = self code text ifFalse: [ 
 		self updateSourceCodeFor: aContext ].
+	formerActionModel := self code interactionModel.
 	self code beForContext: aContext.
+	"add bindings of the old interaction model in the new one"
+	formerActionModel ifNotNil: [ 
+		formerActionModel bindings associations do: [ :assoc | 
+			self code interactionModel addBinding: assoc ] ].
 	self code selectionInterval:
 		(selectionInterval last to: selectionInterval last - 1).
 	self
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -821,6 +821,7 @@ StDebugger >> newDebuggerContext [
 
 	^ self class debuggerContextClass new
 		  exception: self exception;
+		  debugger: self;
 		  yourself
 ]
 
@@ -829,6 +830,7 @@ StDebugger >> newDebuggerContextFor: aContext [
 
 	^ self newDebuggerContext
 		  context: aContext;
+		  debugger: self;
 		  yourself
 ]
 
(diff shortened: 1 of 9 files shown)


src/NewTools-Debugger/StDebuggerExtensionInspectorNodeBuilderVisitor.class.st
```diff
@@ -0,0 +1,41 @@
...
+		  select: [ :aBinding | 
+			  aBinding propertyAt: #comesFromChest ifAbsent: [ false ] ]
+		  thenCollect: [ :chestVariable | 
+			  (StInspectorDynamicNode
+				   hostObject: chestVariable value
+				   label: chestVariable key
+				   value: chestVariable value)
+				  variableTag: 'Chest';
+				  yourself ]
+]
```
StevenCostiou: Should be in Chest as an extension method.

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -1245,14 +1245,19 @@ StDebugger >> updateCodeFromContext [
...
 	self recordUnsavedCodeChanges.
 	selectionInterval := self selectedCodeRangeForContext: aContext.
 	aContext sourceCode = self code text ifFalse: [ 
 		self updateSourceCodeFor: aContext ].
+	formerActionModel := self code interactionModel.
 	self code beForContext: aContext.
+	"add bindings of the old interaction model in the new one"
+	formerActionModel ifNotNil: [ 
+		formerActionModel bindings associations do: [ :assoc | 
+			self code interactionModel addBinding: assoc ] ].
```
StevenCostiou: Review naming it is unclear

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -1245,14 +1245,19 @@ StDebugger >> updateCodeFromContext [
 { #category : #'presenter - code' }
 StDebugger >> updateCodeFromContext: aContext [
 
-	| selectionInterval |
+	| selectionInterval formerActionModel |
 	aContext ifNil: [ ^ self clearCode ].
```
StevenCostiou, line 1252: Add flag: 'to refactor later'?
Ugly and unclear code.

src/NewTools-Debugger/StDebuggerExtensionInspectorNodeBuilderVisitor.class.st
```diff
@@ -0,0 +1,41 @@
...
+	^ #(  ) asOrderedCollection
+]
+
+{ #category : #visiting }
+StDebuggerExtensionInspectorNodeBuilderVisitor >> visit: aStDebuggerExtension [
+
+	| extensionNodes |
+	extensionNodes := super visit: aStDebuggerExtension.
+	"If the visit hook that is used does nothing, then we return an empty collection"
+	^ extensionNodes == self ifTrue: [ self newCollection ] ifFalse: [ extensionNodes ]
```
StevenCostiou: Prefer to store the results into an inst var rather than returning data for which we cannot predict the type.

StevenCostiou (changes requested):

adri09070:
all done

StevenCostiou (approved):

PR #405, opened 2022-07-27
Fix debugger to update debugger extensions when a context is selected in the stack + fixing bytecode extension

### LINKED ISSUE #403: Debugger extensions are not updated when we select a context in the stack
When a context is selected in the stack, debugger extensions are not updated although they should.

For instance, to reproduce an example of the bug:

1) Activate the bytecode extension (Cmd + O + S > Tools > Debugging > Debugger Extensions > Activate extensions > tick Bytecode)

2) Open a debugger (for example, open a playground and evaluate 1 / 0). On the right, you can see the bytecode program of the context `SmallInteger>>#/`.

3) select the context `UndefinedObject>>#DoIt` in the context stack. The bytecode program that is shown on the right is still the bytecode program of the context `SmallInteger>>#/`

This bug proves that debugger extensions should be updated when a context is selected in the stack

adri09070:
Fixes #403.

In the debugger, when we would select a context in the stack, debugger extensions weren't updated.
In the case of the bytecode debugger extension, this would cause that the displayed bytecodes didn't necessarily match the code of the selected context.

There were 2 reasons to that:

1) Debugger extensions were not updated when a context was selected in the stack. I fixed it so that debugger extensions are updated when a context is selected

```Smalltalk
StDebugger>>#initializeStack
...
stackTable transmitDo: [ :context | 
		stackTable selection isEmpty ifFalse: [ 
			self updateInspectorFromContext: context.
			self updateCodeFromContext: context.
			self expandStackIfLastItemIsSelected.
			self updateWindowTitle ] ]
...
```

2) To provide the list of bytecodes, the bytecode extension uses a `SindarinDebugger` that only manipulates the interrupted context (= topContext). So, no matter what, the bytecode list that was displayed stood for the top context instead of the selected context in the debugger. I fixed it so that the bytecode extension uses the debugger itself instead of a sindarin debugger to build the list of symbolic bytecodes that must be displayed.

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -743,6 +743,7 @@ StDebugger >> initializeStack [
 		stackTable selection isEmpty ifFalse: [ 
 			self updateInspectorFromContext: context.
 			self updateCodeFromContext: context.
+			self updateExtensionsFrom: self session.
 			self expandStackIfLastItemIsSelected.
 			self updateWindowTitle ] ].
 	stackHeader := self instantiate: StHeaderBar.
--- a/src/NewTools-Sindarin-Tools/SindarinDebugger.extension.st
+++ b/src/NewTools-Sindarin-Tools/SindarinDebugger.extension.st
@@ -1,13 +1,13 @@
 Extension { #name : #SindarinDebugger }
 
 { #category : #'*NewTools-Sindarin-Tools' }
-SindarinDebugger >> debuggerExtensionKey [
-	^self class debuggerExtensionKey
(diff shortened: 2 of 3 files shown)


StevenCostiou (commented):
@adri09070 For me this is good.
Can you check and confirm that, with the existing extensions, there is no recursive loop between the stack update and the extension update (*i.e.*, stack update triggers extensions update and one of the extensions triggers the stack update again)?

adri09070:
@StevenCostiou 

> Can you check and confirm that, with the existing extensions, there is no recursive loop between the stack update and the extension update (i.e., stack update triggers extensions update and one of the extensions triggers the stack update again)?

I've checked and there shouldn't be any problem

PR #407, opened 2022-08-01
Use stateless compiler API

dionisiydk:
Use stateless compiler API. 
The API was always there (evaluate: source vs source: source; evaluate). So that it's fully compatible any compiler version


dionisiydk:
I will close it for now because the diff is confusing

PR #408, opened 2022-08-09
Added skip command to the expert debugger toolbar command group

StevenCostiou:
Instead of executing a normal *step*, this command skips the next bytecode to be executed.
It has the same granularity than the step-**in** command.
Executing a skip sets the next bytecode in line to be executed, therefore you must be careful when skipping instructions as you could skip a bytecode that sets a required value to execute the rest of your method.

A good example of using skip would be the following:
You start debugging a method in which you write in to a stream.
You notice a problem so you fix your method and recompiles it, which triggers a context restart.
In that case, the execution counter is set back to the beggining of that method but the state is not set back. 
Therefore, if you execute again you will write the same values in your stream.
You can use skip to avoid all writings in your stream until you get back to the correct point of the execution where you can start writing again in the stream.

Available from the expert toolbar sub-menu:
Capture d’écran 2022-08-09 à 15 50 55

and from the code pane contextual menu:
Capture d’écran 2022-08-09 à 15 51 09

### DIFF
--- a/src/NewTools-Sindarin-Commands/SindarinSkipCommand.class.st
+++ b/src/NewTools-Sindarin-Commands/SindarinSkipCommand.class.st
@@ -1,0 +1,27 @@
+Class {
+	#name : #SindarinSkipCommand,
+	#superclass : #SindarinSkipAllToSelectionCommand,
+	#category : #'NewTools-Sindarin-Commands'
+}
+
+{ #category : #'accessing - defaults' }
+SindarinSkipCommand class >> defaultDescription [
+	^ 'Skips execution of the next step, then returns debugger control.'
+]
+
+{ #category : #'accessing - defaults' }
+SindarinSkipCommand class >> defaultName [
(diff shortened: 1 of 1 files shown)


StevenCostiou:
@adri09070 @estebanlm @Ducasse

StevenCostiou:
Unrelated test failing?

estebanlm:
no, the test says there is an unused variable at SindarinSkipCommand>>#execute :)

StevenCostiou:
Hm I read it too fast.

StevenCostiou:
So I fixed it, and also added a class comment because, who knows, sometimes it helps ;)

Ducasse:
Oh this is nice!
Thanks a lot. 
I needed because in some logic I remove a file then do something so to repeat I had to recreate the file or to comment the file deletion.

StevenCostiou:
All green

PR #410, opened 2022-08-12
fix closing Spotter by Esc

pavel-krivanek:


pavel-krivanek:
fixes https://github.com/pharo-project/pharo/issues/11511

PR #415, opened 2022-09-06
[Debugger] Improving skipUpTo stability + reintegration in the advanced debugging tools

adri09070:
skipUpTo command was very unstable and could easily make an image freeze. However, it has been improved a lot via a bunch of PRs in https://github.com/pharo-spec/ScriptableDebugger.

The only PR that needs to be integrated to make it completely stable is this one: https://github.com/pharo-spec/ScriptableDebugger/pull/48. Then, I believe that this could be integrated again in the debugger and that's why I make this PR as some people like @labordep and @astares are interested in this feature.

I also improved the command to make it more intuitive as, until now, the user had to select a message send or an assignment to skip up to this instruction, otherwise it would skip everything. I improved it so that putting a caret before a message send or assignment (or anything else) allows the user to skip up to this instruction.

This PR could be merged after https://github.com/pharo-spec/ScriptableDebugger/pull/48 is merged.


adri09070:
It doesn't look like the failing test is related

MarcusDenker (approved):

PR #419, opened 2022-09-20
Adds a setting to change the default playground cache directory.

lopezca:
This code adds back the setting to change the playground's cache directory.

It also removes the TODO flag on the method of `cacheDirectory:`.

I placed the setting under system, but let me know if it needs to be changed.


Ducasse:
Thanks a lot!!!

estebanlm (approved):
is ok, there is already a setting to change the pharo-local directory globaly, but one for just the playground pages were missing (I even let a TODO there!) :)

PR #421, opened 2022-09-30
Fix the UI of the spotter by outlining the input field. Fix pharo/#10993

privat:
The input field is now light gray (or whatever according to the smalltalk theme), so it is obvious that the x cross belongs to the input field (and clears it) and does not belong to the whole modal window (and mistaken to a close button).

Done with @DanielCamSan

### DIFF
--- a/src/NewTools-Morphic-Spotter/StSpotterStyleContributor.class.st
+++ b/src/NewTools-Morphic-Spotter/StSpotterStyleContributor.class.st
@@ -20,7 +20,7 @@ StSpotterStyleContributor >> styleSheetContribution [
 				addPropertyDrawWith: [ :draw | draw color: Color transparent ] ] ];
 		addClass: 'searchInputField' with: [ :class | class 
 			addClass: 'stSpotterSearch' with: [ :spotterClass | spotterClass
-				addPropertyDrawWith: [ :draw | draw color: Color transparent ];
+				addPropertyDrawWith: [ :draw | draw backgroundColor: Smalltalk ui theme lightBackgroundColor ];
 				addPropertyTextWith: [ :text | text drawKeyboardFocus: false ];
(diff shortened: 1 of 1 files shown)


jordanmontt (approved):

Ducasse:
We should not use Smalltalk ui theme lightBackgroundColor but we should ask the stApplication because it is it that should manage the color. Now I do not know if the stApplication provides this (it was the idea) so 
- if stApplication does not we should introduce it probably proposing for now a dictionary key value for the role and the colors (colorFor: #lightBackgroundColor) but we should check how this is interacting with the style of the application (I never looked at the implementation). 
- if it does we should use it. 
@estebanlm ?

Ducasse:
PS: Smalltalk ui theme lightBackgroundColor is a super bad pattern because this is a global variable accessed from everywhere and we want to kill them all.

estebanlm:
can you put a screenshot on how it looks in white and black themes?

Ducasse:
@esteban does stApplication support theme color access?

privat:
I did just realize that this PR is still not merged. I agree that the proposed `Smalltalk ui theme lightBackgroundColor` is not ideal, but it is consistent with the current implementation where `Smalltalk ui theme` is already used in the same method https://github.com/pharo-spec/NewTools/blob/e831ff29bef86e6bf6c82c735cc26f7af24bb694/src/NewTools-Morphic-Spotter/StSpotterStyleContributor.class.st (and likely in other methods of the tool).

Moving to a better theme color approach is the right thing to do, but it could be done more globally in a future PR that tackle the whole tool.

So my current understanding is that the current PR is a (small) improvement that fixes an existing prevalent issue without introducing (new) technical dept.

Ducasse:
I let estaban manage the integration. @estebanlm

MarcusDenker (approved):
I think we should merge this, we can always improve furtber

PR #427, opened 2022-10-28
inspected object name isnt truncated when it have more space

### LINKED ISSUE #426: Inspector: inspected object name is truncate
**Describe the request**
image
There is a lot of available space to display the name of the inspected object. 
**Expected behavior**
It should use that free space ?

DanielCamSan:
fix #426

### DIFF
--- a/src/NewTools-Core/StHeaderBar.class.st
+++ b/src/NewTools-Core/StHeaderBar.class.st
@@ -65,7 +65,7 @@ StHeaderBar >> initializePresenters [
 		beNotHomogeneous;
 		borderWidth: 3;
 		spacing: 3;
-		add: (titleLabel := self newLabel) expand: false;
+		add: (titleLabel := self newLabel) expand: true;
 		add: (shortcutLabel := self newLabel) expand: false;
 		addLast: (toolbar := self newToolbar) expand: false;
 		yourself).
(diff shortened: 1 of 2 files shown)


src/NewTools-Core/StHeaderBar.class.st
```diff
@@ -65,7 +65,7 @@ StHeaderBar >> initializePresenters [
 		beNotHomogeneous;
 		borderWidth: 3;
 		spacing: 3;
-		add: (titleLabel := self newLabel) expand: false;
+		add: (titleLabel := self newLabel) expand: true;
```
jordanmontt, line 68: The expanding property is always true by default. So, there is no need to declare it. You can just say:
`add: (titleLabel := self newLabel);`

src/NewTools-Inspector/StInspectorModel.class.st
```diff
@@ -85,7 +85,7 @@ StInspectorModel >> inspectedObject: anObject [
 { #category : #accessing }
 StInspectorModel >> labelString [
 	
-	^ self descriptionString truncateWithElipsisTo: 30
+	^ self descriptionString
```
jordanmontt, line 88: The problem when we don't truncate the string is that the title is cutted without any dots that indicates that the name continues.

Capture d’écran 2022-11-03 à 09 03 11

jordanmontt (changes requested):

src/NewTools-Inspector/StInspectorModel.class.st
demarey: I also do not think it is a good change. Maybe it would be nice to have a configurable truncation of Strings, depending on the space available to display it. But anyway, it would be another PR.
I will close this one. Feel free to open a new one.

PR #432, opened 2022-11-15
424-Playground-DNU-when-opening-bindings 

### LINKED ISSUE #424: Playground: DNU when opening bindings
StPlaygroundBindingsPresenter>>#contextMenuActions

DNU for: forSpecNamed:

MarcusDenker:
fix using the code in the issue, no idea if this is the right way but it fixes the problem

fixes #424


estebanlm:
uh, this in fact is because there is a missing method in spec, I will fix it there :P

PR #434, opened 2022-11-21
remove the unused <return:> pragma in spotter

MarcusDenker:
remove the return pragma


MarcusDenker:
fixes https://github.com/pharo-spec/Spec/issues/1327

PR #440, opened 2022-11-24
add a Method Source View for Context

MarcusDenker:
Just a simple source view for Context, context is highlighted (nice for blocks, a bit strange for methods)


MarcusDenker:
Untitled

PR #445, opened 2022-11-29
StDebugger-stackSelectionMethodContext-homeMethod

MarcusDenker:
use #homeMethod (it is odd that the method named -Context returns a method, but that is to be checked later)


MarcusDenker:
failing test not related

PR #447, opened 2022-11-29
Fixing fuel out command + making it more user-friendly

### LINKED ISSUE #446: "Fuel out stack" command does not work in the debugger
This command is supposed to serialize the context stack into a file, to be then materialize later as a post-mortem stack.

How to reproduce the bug:

* Open a playground (CMD + O + W)
* Evaluate `1 / 0` in the playground (CMD + D) to open a debugger
* Right-click on the top context in the context stack and click on Stack > Fuel out stack
* An exception occurs and the stack isn't serialized although it should be:

image

adri09070:
Fixes #446 with help of @ELePors

`StFuelOutStackCommand` was not working as the message `#newDefault` was sent to an instance of `FLDebuggerStackSerializer` which didn't understand the message.

To fix it, we send the message `new` instead of `newDefault` as it is not needed.

Plus, we made the command more user-friendly by using the UIManager to inform the user when the serialization has begun, when it has finished and in what file it has been serialized.

Actually, doing that works but ONLY for one serialization. We manage to serialize the first serialization but the file resulting from the second one is empty and I get weird errors after the second one:

image

We also tried to fork the serialization process so that it doesn't freeze the process, but I removed it because when we tried to do that, we manage to serialize but we get errors when we materialize:

image


StevenCostiou:
What is the status on this one?
If we remove the fork thing, can it be merged?

adri09070:
I'm not sure. I thought that it could be merged without the fork thing, but I had the impression that something else is breaking the command in what we have done.

This is difficult to do tests because the commands breaks each time we change its source code after using it !

adri09070:
> Plus, we made the command more user-friendly by using the UIManager to inform the user when the serialization has begun, when it has finished and in what file it has been serialized.
> 
> Actually, doing that works but ONLY for one serialization. We manage to serialize the first serialization but the file resulting from the second one is empty and I get weird errors after the second one:

I don't get the error anymore in a fresh image. This error could be due to changing the code of the Fuel out command after fueling out once.

> We also tried to fork the serialization process so that it doesn't freeze the process, but I removed it because when we tried to do that, we manage to serialize but we get errors when we materialize:
> 
> image

This remains a mystery. I don't get why forking the process would break the command. So I removed it for now and it can be integrated as it is.

Parallelizing the command could be the object of a second PR

StevenCostiou:
There are conflicts it seems?

adri09070:
Conflict resolved

PR #458, opened 2023-02-03
Debugger-Actions are legacy code from the previous debugger.

jecisc:
This seems to not be used anymore. I propose to remove it to be able to unload this legacy package from Pharo


jecisc:
error unrelated

PR #463, opened 2023-02-10
Removing the TStDebuggerExtension trait from the dummy debugger presenter

### LINKED ISSUE #404: `StDummyDebuggerPresenter` can be activated as a debugger extension but it shouldn't
`StDummyDebuggerPresenter` is a debugger extension that is used to test the debugger extension mechanism. It is also used to test the debugging command mechanism without relying on the `StDebugger`.

This debugger extension can be activated in the Pharo debugging settings (its name is `Test extendedDebuggingToolTitle` in the settings). As no layout is defined for this extension, the `StDebugger` never opens when we try to open a debugger. 

This debugger extension should not be displayed in the Pharo settings

StevenCostiou:
Avoids having this dummy debugger presenter intended for tests as a possible system extension.
Added methods to fill the missing behavior due to the trait removal.

Fixes https://github.com/pharo-spec/NewTools/issues/404

### DIFF
--- a/src/NewTools-Debugger-Tests/StDummyDebuggerPresenter.class.st
+++ b/src/NewTools-Debugger-Tests/StDummyDebuggerPresenter.class.st
@@ -6,10 +6,9 @@ The debugger interface itself is tested in the debugger test class.
 Class {
 	#name : #StDummyDebuggerPresenter,
 	#superclass : #SpPresenter,
-	#traits : 'TStDebuggerExtension',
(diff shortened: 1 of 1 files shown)


adri09070 (approved):

StevenCostiou:
Unrelated failing test?
I see that one often, is this a problem in the base image?

adri09070:
> I see that one often, is this a problem in the base image?

Yes it is

PR #464, opened 2023-02-10
Adding new rewrite tools

jordanmontt:
This PR adds the Rewriter Tools. Also has two backends for doing the transformations. One can use either `RB` or `Renraku`

It has the rewriter
Capture d’écran 2023-02-10 à 17 16 55

Scope selector that allows to apply the transformations to custom classes
Capture d’écran 2023-02-10 à 17 17 18

Match Tool 
Capture d’écran 2023-02-10 à 17 13 10

Expression Browser
Capture d’écran 2023-02-10 à 17 13 46

Ocurrences Browser
Capture d’écran 2023-02-10 à 17 14 59

Example rules ready to be used
Capture d’écran 2023-02-10 à 17 15 23

Explanation about Rewrite Rules
Capture d’écran 2023-02-10 à 17 13 21

### DIFF
--- a/src/BaselineOfNewTools/BaselineOfNewTools.class.st
+++ b/src/BaselineOfNewTools/BaselineOfNewTools.class.st
@@ -30,7 +30,7 @@ BaselineOfNewTools >> baseline: spec [
 			package: 'NewTools-Inspector-Extensions' with: [ spec requires: #('NewTools-Core') ];
 			package: 'NewTools-Inspector-Tests' with: [ spec requires: #('NewTools-Inspector') ];
 			"debugger"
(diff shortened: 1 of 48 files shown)


estebanlm:
PR is ok. 
but beware! 
Including it here does not means it will be loaded in Pharo. You need an extra step for it :)

jordanmontt:
Which is that step? :)

I already edited the baseline to include the tools

estebanlm:
the baseline of Pharo :)

jordanmontt:
Nop, I just tested and it is included in Pharo

PR #466, opened 2023-02-17
FIx protocols

jecisc:


jecisc:
The failing release test is also present in Pharo (and the problem is in Pharo and not NewTools) and I’ll fix it this afternoon

PR #467, opened 2023-02-21
Debugger configuration toolbar improvements

StevenCostiou:
This PR improves the design, test coverage and extensibility of the debugger configuration toolbar situated above the stack presenter.
Before, this toolbar was initialized by creating and adding commands directly when the toolbar was created, during the `StDebugger` presenter initialization.
Because of that, it was:
- not tested
- not extensible without modifying the code of the debugger
- making the code poorly readable

This PR uses a dedicated class to build the same configuration command tree that is added to the debugger root command tree. 
Using spec presenters extensibility mechanisms, it is now possible to complete this tree with new command, improving the extensibility of the debugger.
The code is also cleaner.

Note that part of the unit tests are there to enforce the specification of the debugger features.
They look trivial (in part), but they are there to tell you when a feature is modified by a code change and is not anymore conform to the debugger design (in which case, either the design evolves or it is a bug and you need to fix it).


StevenCostiou:
Everything looks good.

Ducasse:
This is good to see all these improvements.

src/NewTools-Debugger/StDebuggerConfigurationToolbarCommandTreeBuilder.class.st
```diff
@@ -0,0 +1,46 @@
...
+{ #category : #accessing }
+StDebuggerConfigurationToolbarCommandTreeBuilder class >> groupName [ 	
+	^'Configuration'
+]
+
+{ #category : #building }
+StDebuggerConfigurationToolbarCommandTreeBuilder >> buildDebuggerCommandGroup [
+
+	| group |
+	self group.
```
adri09070: this line looks useless (same as line 27 without assignment)

src/NewTools-Debugger/StDebuggerConfigurationToolbarCommandTreeBuilder.class.st
```diff
@@ -0,0 +1,46 @@
...
+		group register: specCommand ].
+	^ group
+]
+
+{ #category : #'accessing - commands classes' }
+StDebuggerConfigurationToolbarCommandTreeBuilder >> configurationCommandClasses [
+
+	^ {
+		  StDebuggerToggleFilterStackCommand.
+		  StDebuggerSettingsCommand }
```
adri09070: The classes shouldn't be hardcoded. So, to extend the debugger command classes, we need to change this collection? 

Instead of changing the code of the debugger, we would need to change the collection in this method. It just moves the problem elsewhere no?

Why don't we use a trait (or an abstract class) to make this method return `TStDebuggerConfigurationCommand users` (or `StDebuggerConfigurationCommand subclasses`)?

src/NewTools-Debugger-Tests/StDebuggerConfigurationCommandTreeBuilderTest.class.st
```diff
@@ -0,0 +1,47 @@
...
+		withContext: debugger
+		displayStrategy: CmUIDisplayAsGroup
+]
+
+{ #category : #tests }
+StDebuggerConfigurationCommandTreeBuilderTest >> testConfigurationCommandsClasses [
+
+	self assertCollection: builder configurationCommandClasses equals: {
+			StDebuggerToggleFilterStackCommand.
+			StDebuggerSettingsCommand }
```
adri09070, line 29: So, if we add a class, the test should break ? Looks weird

src/NewTools-Debugger-Commands/StDebuggerSettingsCommand.class.st
```diff
@@ -12,11 +12,22 @@ StDebuggerSettingsCommand class >> defaultDescription [
...
+]
+
 { #category : #default }
 StDebuggerSettingsCommand class >> defaultName [ 
 	^'Settings'
 ]
 
+{ #category : #initialization }
+StDebuggerSettingsCommand class >> defaultShortcut [
+	self flag: 'Any suggestion for a "filter stack" shortcut?'.
```
adri09070: $f meta + $s ?

src/NewTools-Debugger-Commands/StDebuggerToggleFilterStackCommand.class.st
```diff
@@ -12,11 +12,22 @@ StDebuggerToggleFilterStackCommand class >> defaultDescription [
...
+]
+
 { #category : #default }
 StDebuggerToggleFilterStackCommand class >> defaultName [ 
 	^'Toggle stack filtering'
 ]
 
+{ #category : #initialization }
+StDebuggerToggleFilterStackCommand class >> defaultShortcut [
+	self flag: 'Any suggestion for a "configuration" shortcut?'.
```
adri09070: $c meta + $f ?

src/NewTools-Debugger/StDebuggerConfigurationCommandTreeBuilder.class.st
```diff
@@ -1,3 +1,11 @@
+"
+I build the default command tree for the debugger configuration.
+The command classes I instantiate are specified under the ""accessing - commands classes"" protocol.
+The group names and descriptions are specified under the ""accessing"" protocol.
+
+You can evolve the default groups names and descriptions and the default command classes, but you cannot extend me.
```
adri09070: what do you mean with this sentence ?

src/NewTools-Debugger-Tests/StDebuggerConfigurationCommandTreeBuilderTest.class.st
StevenCostiou: Yes, the test is a specification, it is meant to tell what the default debugger commands are.
It means that if we change the default commands, we get notified of it because the test breaks.
We then have to decide if that is a bug (then fix it) or if the command should change on purpose and therefore update the test.

It is not elegant, but it facilitates maintenance compared to using pragmas as in the previous implementation.
The alternative using pragmas would be to have lots of pragmas, each of them representing a particular group of commands.

adri09070: I think we should differentiate defaultConfigurationCommandClasses from configurationCommandClasses (cf. my other comment above).

You could then have a test that checks that defaultConfigurationCommandClasses are equals to this collection and another test that checks that configurationCommandClasses contain the default ones.

src/NewTools-Debugger/StDebuggerConfigurationToolbarCommandTreeBuilder.class.st
StevenCostiou: My proposition is to avoid at all costs to use another entity to recover the commands (such as a trait, abstract class, or worse, pragmas).
This leads to maintenance hell because it becomes extremely tedious to track where and when during an execution the commands are instantiated and stored in a menu.

The design choice here is to have these `CommandTreeBuilder` classes, that indeed hardcode the default debugger command tree.
However, that is just how the debugger is **configured by default**, and not how it should be extended.
To extend the **default** debugger commands, we have to use the SpPresenter extension mechanism, that does not require to change the debugger code, or those tree builder code, or anything in the default Pharo image.
The spec mechanism gives access to that default command tree, from which we can remove, modify or add commands.

Therefore:
- To modify the debugger default commands, just modify these classes. It is ok, because they statically represent the commands the debugger should expose by default. Maintenance is easier, because if we need to track one of the default commands when they are added to the debugger, we can find them statically using the system tools.
- To extend or customize the debugger default commands, use the default spec extension mechanism.

src/NewTools-Debugger/StDebuggerConfigurationCommandTreeBuilder.class.st
StevenCostiou: Poor phrasing, I meant: the commands statically configured in the class can be modified there, but by default the debugger command tree builders are not meant to be specialized through subclassing.
I will update the comment.

src/NewTools-Debugger/StDebuggerConfigurationToolbarCommandTreeBuilder.class.st
StevenCostiou: I do not have this in my image, actually that class was removed I think.
I will check.

StevenCostiou: Yes, it was removed by https://github.com/pharo-spec/NewTools/pull/467/commits/9a9e2564ea92afd24be6e8dd044d111e6505ff59

StevenCostiou:
I did the shortcuts update and the class comment fixes.

src/NewTools-Debugger/StDebuggerConfigurationToolbarCommandTreeBuilder.class.st
adri09070: Ok, I just realized we can use `<extensionCommands>` in `StDebugger class`

adri09070 (approved):

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -711,6 +718,9 @@ StDebugger >> initializeShortcuts: aWindowPresenter [
 StDebugger >> initializeStack [
 
 	stackTable := self newTable.
+	stackTable 
+		activateOnDoubleClick;
+		whenActivatedDo: [ :selection | self doBrowseClass ].
```
adri09070, line 723: This shouldn't be related to this PR.

(I don't manage to find the related issue about that)

PR #470, opened 2023-02-28
Do not ask the user a method protocol first but instead try to auto classify the method and only ask the user if auto classification failed

demarey:
Do not ask the user a method protocol first but instead try to auto classify the method and only ask the user if auto classification failed


demarey:
WARNING: This PR is dependent on https://github.com/pharo-project/pharo/pull/12882

jecisc:
Pharo PR was integrated

PR #472, opened 2023-03-01
Fixing debugger inspector that wasn't updating when the debugger entered a new optimized scope

### LINKED ISSUE #450: Debugger inspector doesn't update when entering inlined blocks, which hides its temporaries
# How to reproduce the bug:

- In a playground, copy/paste this code and debug it:
```Smalltalk
| a b|
a := 1.
true ifTrue: [ | d | d := 2. false ]
```
- The debugger opens and has not entered the inlined block `ifTrue:` yet. Perform stepOver and the block is entered. Then the variable `d` should be shown by the inspector but it isn't.
- 
image

- If you select another context in the context stack and then choose the DoIt context again, then the variable `d` will appear in the debugger inspector, which proves that the debugger inspector doesn't update when it enters inlined blocks.

image

# Notes:

- If you debug in a playground the following code, the inspector show the variables from the inlined block because the debugger opens when it has already entered the inlined block, as there's no instruction before. This supports the hypothesis that the debugger inspector not being updated is the problem.

```Smalltalk
| a b|
a := 1.
true ifTrue: [ | d | d := 2. false ]
```

- This bug can be reproduced at least in Pharo 11 and Pharo 10 but I wouldn't be surprised if it could be reproduced with previous versions of Pharo.

adri09070:
Fixes #450 .

When the debugger entered/left an inlined block with a step, the debugger inspector wouldn't update. As a consequence, if new temporaries were declared in this optimized scope, their values wouldn't be displayed in the inspector (unless this context was deselected and then selected again in the stack).

To fix that, in the debugger action model, before performing a step I store the AST scope of the context we step into/over/through.
Then, after updating the stack of the `StDebugger`, I compare if the interrupted context after stepping is still the context that was selected before stepping. If this is the case, I ask the raw inspection presenter to update its nodes from the old AST scope with the nodes from the current AST scope.
To do that, I use methods on collections to remove all temp nodes that were in the old scope and that aren't anymore in the new scope (in case we leave an inlined block) and I also add all temp nodes that are in the new scope and that were not in the old scope (in case we enter an inlined block).
Using methods on collections allows to do nothing in case both scopes are the same or in case both scopes aren't the same but have the same temps.

### DIFF
--- a/src/NewTools-Debugger-Tests/StDebuggerTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerTest.class.st
@@ -294,6 +294,45 @@ StDebuggerTest >> testContextTempVarList [
 ]
 
 { #category : #'tests - context inspector' }
+StDebuggerTest >> testContextTempVarListUpdatesTempsWhenEnteringOrLeavingInlinedBlocks [
+
+	| contextItems inspectorTable newContextItems |
+	dbg := self debuggerOn: session.
+	inspectorTable := dbg inspector getRawInspectorPresenterOrNil
+		                  attributeTable.
+
+	"We stop on the node `i = 1` node"
+	3 timesRepeat: [ dbg stepOver ].
+
+	contextItems := inspectorTable roots copy.
+	self assert: contextItems size equals: 8.
+
+	"We enter inlined block"
+	dbg stepOver.
+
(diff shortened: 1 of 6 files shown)


StevenCostiou (approved):
Changes look good and tests are passing-shot.

PR #473, opened 2023-03-03
Fixing Chest's group that should be loaded in the baseline

adri09070:
Right now, nothing is loaded because Chest doesn't specify anything for the 'default' group.

This will load  `Chest`, `Chest-Commands` and `Chest-Tests`


adri09070:
Still, nothing is loaded... I try to understand why

StevenCostiou (approved):
Looks good.
I will merge it if the CI passes.

adri09070:
But it's still not good. It still doesn't load Chest and I don't understand why

StevenCostiou:
Can you try removing the 'default' group loading?

adri09070:
It doesn't change anything.

It may be that it is not loaded because nothing that is loaded by default has a dependency to Chest

adri09070:
The only difference I see with Sindarin, which is loaded, is that NewTools-Sindarin-Commands requires the package `Sindarin` and is included in the Debugger group, which is loaded by default

adri09070:
Ok, now it works

adri09070:
Failing test looks unrelated

MarcusDenker:
I looks like it breaks the bootrap (the full image load):

https://github.com/pharo-project/pharo/issues/12926

PR #481, opened 2023-03-09
Refactoring debugger extension system to update extensions in all debuggers dynamically, when an extension is (un)activated

adri09070:
When we used to activate debugger extensions in the pharo settings, if a debugger was already opened, we had to close the debugger and open a new one and step to the point we were before closing the first debugger...

This was tedious.

I fixed that so that all debuggers update automatically with their dynamic spec layout, each time an extension is enabled/disabled.

It works, but I put it as a draft because :

- I would like to write unit tests (but sadly, I'm never really inspired when it comes to writing UI unit tests)
- the order of the extensions after update is inversed, and I have no idea of why it does that


adri09070:
I couldn't manage to keep the extension display order, although I tried to sort the pages by 100,000 different ways in 100,000 different places in the code. So, I'll leave it as it is. It seems to be a `SpNotebookPresenter` problem

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -231,8 +228,7 @@ StDebugger class >> taskbarIconName [
 { #category : #'tools registry' }
 StDebugger class >> usesExtensions [
 
-	^ self activateDebuggerExtensions and: [ 
-		  self hasAnyActivatedExtension: self extensionToolsClasses ]
+	^ self hasAnyActivatedExtension: self extensionToolsClasses
 ]
```
StevenCostiou, line 232: I wonder now if that method is necessary with the new refactoring.
Is it used/called somewhere else?

src/NewTools-Debugger-Tests/TStDebuggerExtensionTest.class.st
```diff
@@ -52,3 +53,18 @@ TStDebuggerExtensionTest >> testShowInDebugger [
...
+	listener := StDebuggerExtensionTestListener new.
+	SystemAnnouncer uniqueInstance weak
+		when: StDebuggerExtensionActivationToggle
+		send: #tag:
+		to: listener.
+		
+	self assert: listener tag isNil.
+	traitUser class showInDebugger: true.
+	self assert: listener tag class equals: StDebuggerExtensionActivationToggle 
+]
```
StevenCostiou: This test seems to test that the system annoncer correctly sends the new announcement. 
I would say it is not related to the debugger changes, and should be removed.
Such test probably already exist in the announcement framework test package.
What do you think?

src/NewTools-Debugger-Tests/StDebuggerExtensionMechanismTest.class.st
```diff
@@ -77,11 +131,22 @@ StDebuggerExtensionMechanismTest >> testRegisterExtensionTool [
...
-	self assert: self debuggerClass usesExtensions
+	self debuggerClass extensionToolsClasses ifNotEmpty: [
+		| activatedExtension |
+		activatedExtension := self debuggerClass extensionToolsClasses
+			                      asOrderedCollection first.
+		activatedExtension showInDebugger: true.
+		self
+			assertCollection: self debuggerClass activatedDebuggerExtensions
+			hasSameElements: { activatedExtension }.
+		self assert: self debuggerClass usesExtensions ]
```
StevenCostiou: Written like this, the test might pass without testing anything because it depends on the fact that the debugger class has extension tool classes. If it is empty (line 143) then everything below will not be tested.

Is it possible to rewrite the test in a way that we ensure everything below line 143 is executed?

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -629,13 +640,18 @@ StDebugger >> initialExtent [
...
+	self debuggerActionModel updateContextPredicate.
 	self setSessionHolderSubscriptions.
 	self forceSessionUpdate.
 	self subscribeToMethodAddedAnnouncement.
-	
-	programmaticallyClosed := false.
+	SystemAnnouncer uniqueInstance weak
+		when: StDebuggerExtensionActivationToggle
+		send: #updateExtensionsFromAnnouncement:
+		to: self.
```
StevenCostiou: This is good, however would it make sense to extract that into a specific method?
Something like `registerToSystemAnnouncements`? 
It could make cleaner and clearer code, where we call initialize methods of different nature/concerns instead of flattening these concerns into the initialize method.
What do you think?

src/NewTools-Debugger-Tests/StDebuggerExtensionMechanismTest.class.st
```diff
@@ -22,34 +22,88 @@ StDebuggerExtensionMechanismTest >> debuggerClass [
...
 	| testClass |
 	testClass := StDebuggerExtensionGeneratorForTests testClass.
-	self deny: (self debuggerClass hasAnyActivatedExtension: { testClass }).
+	self deny:
+		(self debuggerClass hasAnyActivatedExtension: { testClass }).
 	testClass showInDebugger: true.
-	self assert: (self debuggerClass hasAnyActivatedExtension: { testClass })
+	self assert:
+		(self debuggerClass hasAnyActivatedExtension: { testClass }).
+	testClass showInDebugger: false
```
StevenCostiou: What if there is a failure before we get to the last line that puts the show in debugger boolean to false?
Could it leave that boolean to true?
In that case, it should be in the tearDown to ensure that it is called, except if the test class is genreated in the setup then deleted in the tear down.

StevenCostiou (changes requested):
I mostly have questions, that when answered (or after actions are taken if necessary) will lead to the approval of the PR.

src/NewTools-Debugger-Tests/TStDebuggerExtensionTest.class.st
adri09070: Yes, the test is weird.

We did that when we pair-programmed 10 months ago

src/NewTools-Debugger-Tests/StDebuggerExtensionMechanismTest.class.st
adri09070: Yes, it should be set to false as I do it for the StDummyDebuggerPresenter.


What happens though is that if you debug this test (or actually anything else while the test is running), an empty extension will appear on the right of all debuggers that are currently open and it will disappear when the test is entirely executed

src/NewTools-Debugger/StDebugger.class.st
adri09070: Are you talking about `usesExtension` or `hasAnyActivatedExtension` ?

adri09070: `usesExtensions` is used in 
```Smalltalk
StDebugger>>#setStackAndCodeContainer

	stackAndCodeContainer := self class usesExtensions
		                         ifTrue: [ 
		                         self stackAndCodeWithExtensionsLayout ]
		                         ifFalse: [ self stackAndCodeLayout ]```
and it could be replaced by 
```Smalltalk
self extensionTools isEmpty
```
that could be extracted as a `usesExtensions` method on instance side

adri09070: It is also used in `StDebugger>>#initializeExtensionTools` but we could remove the call as the rest of the method would work with empty collections if no extension is activated

src/NewTools-Debugger-Tests/StDebuggerExtensionMechanismTest.class.st
adri09070: I don't know. I don't think it's possible and that's why we do this.

Because we can't ensure that there is a debugger extension in the image.

And we can't use any debugger extension for test because `StDebugger class>>#activatedDebuggerExtensions` rejects them.

Initially the test was about activating/unactivating the `activateDebuggerExtensions` setting, which doesn't exist anymore. Maybe the test should just be removed.

Or, we need to generate an extension class on the fly. But it needs to inherit from `SpPresenter`, to define a layout and the explicit requirements from `TStDebuggerExtension`. I haven't managed to do it and that's why I created the debugger extension for tests, which was previously generated.
It is possible that I don't know how to compile methods correctly though

src/NewTools-Debugger/StDebugger.class.st
StevenCostiou: Ok, my bad.
For me it is ok.

adri09070:
The failing test is due to system subscriptions. I don't know how to test that the new debugger is in the subscribers of the SystemAnnouncer. 
I test that the last subscriber is the new debugger but it appears that it's not always the case ? I don't manage to get my test failing in my image.

I will try to do a more generic test

StevenCostiou:
There is a failing test:
```
#########################
# 1 tests did not pass: #
#########################

StDebuggerTest
 ✗ #testUpdateExtensionSubscription (44ms)
TestFailure: a StDebugger is not identical to a StDebugger.
StDebuggerTest(TestAsserter)>>assert:description:resumable:
StDebuggerTest(TestAsserter)>>assert:description:
StDebuggerTest(TestAsserter)>>assert:identicalTo:
StDebuggerTest>>testUpdateExtensionSubscription ...assert: announcementSubscriptions last subscriber
		identicalTo: dbg
StDebuggerTest(TestCase)>>performTest
```

adri09070:
Something looks completely  broken and I'm not really sure if it is my fault or not

adri09070:
I restarted the CI is green.

I don't know what was wrong with the previous CI run

StevenCostiou (approved):

PR #487, opened 2023-03-16
fixing previousASTScope after recompiling a method from the debugger,  from inside a block

### LINKED ISSUE #486: Debugger: `#<` was sent to nil when recompiling a method in the debugger, when we are inside a block
This is because when we recompile a method in the debugger while being inside a block, contexts are popped to restart the method context.
The debugger action model tries to get an AST scope from the block context that is thus terminated. That's why the block context pc is nil:

image

image

adri09070:
Fixes #486 

When we recompile a method in the debugger while being inside a block, contexts are popped to restart the method context.

The debugger action model tries to get an AST scope from the context we're trying to rewind (i.e: the block context) that is thus terminated. That's why the block context's pc was nil.

To fix that, I get the previousASTScope from the method context because that's the one that I need in order to avoid a faulty inspector update

### DIFF
--- a/src/NewTools-Debugger/StDebuggerActionModel.class.st
+++ b/src/NewTools-Debugger/StDebuggerActionModel.class.st
@@ -285,11 +285,19 @@ StDebuggerActionModel >> proceedDebugSession [
 
 { #category : #actions }
 StDebuggerActionModel >> recompileMethodTo: aString inContext: aContext notifying: aNotifyer [
+
+	| methodContext |
 	aContext ifNil: [ ^ self ].
 	self session
 		recompileMethodTo: aString
 		inContext: aContext
-		notifying: aNotifyer
+		notifying: aNotifyer.
+		
+	methodContext := aContext.
+	[ methodContext outerContext ] whileNotNil: [
+		methodContext := methodContext outerContext ].
+	previousASTScope := (methodContext compiledCode sourceNodeForPC:
+		                     methodContext pc) scope
 ]
 
 { #category : #context }
--- a/src/NewTools-Debugger/StDebuggerActionModel.class.st
+++ b/src/NewTools-Debugger/StDebuggerActionModel.class.st
@@ -285,17 +285,15 @@ StDebuggerActionModel >> proceedDebugSession [
 
 { #category : #actions }
 StDebuggerActionModel >> recompileMethodTo: aString inContext: aContext notifying: aNotifyer [
-	| methodContext |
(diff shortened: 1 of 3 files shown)


StevenCostiou:
Can I ask for a test? :D

adri09070:
I don't know how to test it because when you recompile a method from a block, a pop-up appears to ask the user if they're sure

adri09070:
Needs https://github.com/pharo-project/pharo/pull/13112 to be merged first

StevenCostiou:
All tests passing.
@MarcusDenker could you review and merge?

src/NewTools-Debugger/StDebuggerActionModel.class.st
```diff
@@ -285,14 +285,19 @@ StDebuggerActionModel >> proceedDebugSession [
...
 	self session
 		recompileMethodTo: aString
 		inContext: aContext
 		notifying: aNotifyer.
-	previousASTScope := (aContext compiledCode sourceNodeForPC:
-		                     aContext pc) scope
+		
+	methodContext := aContext.
+	[ methodContext outerContext ] whileNotNil: [
+		methodContext := methodContext outerContext ].
```
MarcusDenker: What I do not like is that we rely on the outerContext, this is no problem now, but if we change to clean blocks (or even just have the outerContext when we have a non-local return), this would stop working. (but it would be ok for the current Pharo11/Pharo12).

If we look at #home, does your loop not do the same? And always returns the home context?

If we want the home, we might be able to use "aContext activeHome" here... as we are interested in the home context on the stack? If true, this would avoid the need for an otherContext to to be there

adri09070: > What I do not like is that we rely on the outerContext, this is no problem now, but if we change to clean blocks (or even just have the outerContext when we have a non-local return), this would stop working. (but it would be ok for the current Pharo11/Pharo12).
> 
> If we look at #home, does your loop not do the same? And always returns the home context?
> 
> If we want the home, we might be able to use "aContext activeHome" here... as we are interested in the home context on the stack? If true, this would avoid the need for an otherContext to to be there

I think `#activeHome` should work. And with the home, will it work for clean blocks?

MarcusDenker: yes, without clean blocks active, #home would be ok

MarcusDenker (approved):

PR #492, opened 2023-03-18
skip testDoEvaluateDeclareGlobal for now

privat:
This test tries to hijack low-level compiler stuff. This break PR to improve stuff: https://github.com/pharo-project/pharo/pull/13047

So disable it for now, we will try to enable it back (or in a better form) during Pharo 12 dev cycle (we are very early in the cycle, so it's doable).


jecisc (approved):
The compiler cleanup are really nice, it's worth to disable this test for a while (especially since it does not impact P11)

PR #493, opened 2023-03-18
skip another testDoEvaluateDeclareGlobal for now

privat:
There were two of them! (aren't clone supposed to be bad or something?)

See #492 for the first one


jecisc (approved):
review: Same as previous PR

PR #495, opened 2023-03-23
Fix glitch on method creation

### LINKED ISSUE #489: Debugger: Creating an unexisting method does not refresh the view when stack is unfiltered

jecisc:
When creating a new method through the debugger, the code presenter is not reset, leading to the debugger showing a doesNotUndestard: instead of the new method. 

Fixes #489


jecisc:
This got another PR just before 
Oups
https://github.com/pharo-spec/NewTools/pull/491

Since I don't know the debugger much, the other fix is probably better

PR #497, opened 2023-03-27
More inspector extensions

svenvc:
Introduction of SpSimpleInspectorBuilder to aid in the construction of simple key/value inspector extensions that can be written without knowledge of Spec or the inspector implementation.

Add examples for Integer, Float, Character, String and ByteArray


svenvc:
Thx!

PR #506, opened 2023-04-17
StTestDebuggerProvider class>>#compileMissingClassContextBuilder use permitUndeclared:

privat:
This makes the intention explicit and lessen the reliance on current heuristic


StevenCostiou (approved):
The change is ok from the debugger perspective.
Failing tests seems to be a github actions configuration problem.

MarcusDenker (approved):

PR #509, opened 2023-04-19
Added inspector extensions for RPackage

jordanmontt:
Enregistrement-de-lecran-2023-04


jecisc:
The failure does not seems related since it does not introduce any new class

jecisc (approved):

PalumboN (approved):

PR #510, opened 2023-04-20
Adding inspector transmission node

jordanmontt:
Currently there is two solutions for having the transmissions on the inspector

- Either you use StInspectorNode to wrap your data, but it very limited with the transmissions
- Either you create a presenter class and customize the transmissions. Bad side is that you need to create a class and for simple presenters is not convenient.

I was thinking on this the other day. So you are right, the blog post and StInspectorNode are not compatible.

One solution that I can think is to create a StInspectorTransmissionNode that manages the transmission. This can be a third option when you have a small presenter that you want to put in the inspector and you don't want to create a class but you want the power of transmissions.

This PR introduces the class `StInspectorTransmissionNode` that allows to the user to specify custom transmissions inside the Inspector in an easy way.

Let's take this code:

```st
<inspectorPresentationOrder: 1 title: 'Demo'>
    | items |
    items := #( 1 2 3 4 5 6 7 8 ).

    ^ SpListPresenter new
        items: (items collect: [ :e |
            StInspectorTransmissionNode hostObject: e transmissionBlock: [ :theObject | theObject + 100 ] ]);
        display: [ :e | e hostObject ];
        yourself
```

This produces the inspector extension that add 100 when you click on a number 

image

### DIFF
--- a/src/NewTools-Inspector/StInspectorTransmissionNode.class.st
+++ b/src/NewTools-Inspector/StInspectorTransmissionNode.class.st
@@ -1,0 +1,35 @@
+Class {
+	#name : #StInspectorTransmissionNode,
+	#superclass : #StInspectorNode,
+	#instVars : [
+		'transmissionBlock'
+	],
+	#category : #'NewTools-Inspector-Model'
+}
+
+{ #category : #'instance initialization' }
+StInspectorTransmissionNode class >> hostObject: anObject transmissionBlock: aFullBlockClosure [
+
+	^ self new
+		  hostObject: anObject;
+		  transmissionBlock: aFullBlockClosure;
+		  yourself
+]
+
+{ #category : #accessing }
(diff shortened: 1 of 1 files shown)


svenvc:
LGTM

MarcusDenker (approved):

estebanlm:
sorry for coming late to this, but an example in the class comment would be useful. 
also, I do not like `transmissionBlock:` as a name :P 
(I never liked the "type declaring names", `transmissionAction:` or just `transmits:` or similar are better names IMO.

jordanmontt:
Hello @estebanlm 
I will add a new PR with your observations. Maybe tomorrow on the Pharo sprint

jordanmontt:
Related PR: https://github.com/pharo-spec/NewTools/pull/547

PR #513, opened 2023-04-28
Fix DNU when recompiling block with dead home

guillep:
Choose endPC as home pc in that case.
Check that the home is recompiled and the stack does not move

Depends on https://github.com/pharo-project/pharo/pull/13592


adri09070:
I experimented with your branch because I wanted to check some things and I found a bug: 

If you recompile the method with the dead block, it will work the first time but not if you attempt to recompile it a second time in the same debugger.

adri09070 (approved):
The only reason I haven't approved the changes yet is that I'm not sure that the tests actually test the right things, and I wonder if it could not lead to another (minor) bug in the debugger inspector update.

Still, it fixes an annoying bug, so I think we can merge it and see if it actually leads to other bugs

PR #521, opened 2023-05-11
Add stack coloring to the debugger

jecisc:
This PR add some stack coloring to the debugger.

Now the stack has 3 different text colors:
- One when the context of a stack element is from the same class than the selected context
- One when the context of the stack element is from the same package than the selected context
- And the default text color otherwise

This is especially useful when debugging some pattern such as visitors or composites...

image

image

I wanted to add an info popup to describe the feature but I didn't find a way to display text in Spec, just strings.

THIS REQUIRES: https://github.com/pharo-project/pharo/pull/13693

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -737,39 +737,38 @@ StDebugger >> initializeShortcuts: aWindowPresenter [
 StDebugger >> initializeStack [
 
 	stackTable := self newTable.
-	stackTable 
+	stackTable
 		activateOnDoubleClick;
 		whenActivatedDo: [ :selection | self doBrowseClass ].
 	stackTable
-		addColumn: ((SpImageTableColumn
-				  title: ''
-				  evaluated: [ :context | self stackIconForContext: context ])
+		addColumn: ((SpImageTableColumn title: '' evaluated: [ :context | self stackIconForContext: context ])
 				 width: 8;
 				 sortFunction: nil;
 				 yourself);
-		addColumn: ((SpStringTableColumn
-				  title: 'Class'
-				  evaluated: [ :context | self printReceiverClassInContext: context ])
+		addColumn: ((SpStringTableColumn title: 'Class' evaluated: [ :context | self printReceiverClassInContext: context ])
+				 displayColor: [ :context | self stackColorForContext: context ];
 				 sortFunction: nil;
 				 yourself);
-		addColumn:
-			((SpStringTableColumn title: 'Method' evaluated: [ :context | 
+		addColumn: ((SpStringTableColumn title: 'Method' evaluated: [ :context |
 					  | method |
 					  method := context method.
 					  method isCompiledBlock
 						  ifTrue: [ method sourceNode sourceCode ]
 						  ifFalse: [ method selector ] ])
+				 displayColor: [ :context | self stackColorForContext: context ];
 				 sortFunction: nil;
 				 yourself);
-		addColumn:
-			((SpStringTableColumn title: 'Package' evaluated: [ :context | 
+		addColumn: ((SpStringTableColumn title: 'Package' evaluated: [ :context |
 					  | package |
 					  package := context method package.
-					  package ifNil: [ '-' ] ifNotNil: [ package name asString ] ])
+					  package
+						  ifNil: [ '-' ]
+						  ifNotNil: [ package name asString ] ])
+				 displayColor: [ :context | self stackColorForContext: context ];
 				 sortFunction: nil;
 				 yourself).
-	stackTable transmitDo: [ :context | 
-		stackTable selection isEmpty ifFalse: [ 
+	stackTable transmitDo: [ :context |
+		stackTable selection isEmpty ifFalse: [
 			self updateInspectorFromContext: context.
 			self updateCodeFromContext: context.
 			self updateExtensionsFrom: self session.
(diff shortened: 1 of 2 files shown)


StevenCostiou (commented):
Changes are ok.
There are tests failing because of the changes though.

jecisc:
This is because the Pharo PR has not been reviewed and integrated yet. I’ll restart it when it will be the case.

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -1153,6 +1166,22 @@ StDebugger >> stackAndCodeWithExtensionsLayout [
...
+	
+	If they are from the same class, we return a certain 'primary' highlight color.
+	If they are from the same package (but different class), we return a 'secondary' highlight color.
+	Else, we just keep the normal style."
+
+	^ (self isInSelectedContextClass: context)
+		  ifTrue: [ self theme highlightTextColor ]
+		  ifFalse: [
+			  (self isInSelectedContextPackage: context)
+				  ifTrue: [ self theme highlightTextColor alpha: 0.5 ]
```
privat: According to screenshot, using alpha here make this second-class highlightTextColor **less** visible than the classic text color as it blends with the backgound.

Maybe add an explicit new color entry `highlightTextColorLevel2` or dynamically mix highlightTextColor and textColor together

jecisc: image

Is that better?

StevenCostiou:
Would it be possible to have a setting to (de)activate the coloring?
I guess the next step (after that PR) would be to have a system to custsomize the coloring? :P

jecisc:
I can check to do that when I'll have the time

For the customization, I made it part of the theme. So user can create their own theme. We can also see to improve the default values Pharo's theme if someone has knowledge on color palette.

Another idea would be to have in the setting the possibility to customize all colors of the theme but this is out of the scope of this feature :)

StevenCostiou:
For the customization, I meant customizing what to highlight and under what conditions.
From there why not choosing the color from their application or directly in their customization.

jecisc:
> Would it be possible to have a setting to (de)activate the coloring? I guess the next step (after that PR) would be to have a system to custsomize the coloring? :P

Done!

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -144,12 +145,37 @@ StDebugger class >> debuggerContextClass [
 	^ StDebuggerContext
 ]
 
+{ #category : #accessing }
+StDebugger class >> debuggerSettingsOn: aBuilder [
+
```
StevenCostiou: could we call it `stackColoringSettingOn:`?
Because there are other debugger settings than this one.

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -144,12 +145,37 @@ StDebugger class >> debuggerContextClass [
 	^ StDebuggerContext
 ]
 
+{ #category : #accessing }
```
StevenCostiou: In the debugger the settings should be under the `settings` protocol

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -144,12 +145,37 @@ StDebugger class >> debuggerContextClass [
...
+StDebugger class >> debuggerSettingsOn: aBuilder [
+
+	<systemsettings>
+		(aBuilder setting: #enableStackColoring)
+				label: 'Enable stack coloring';
+				target: StDebugger;
+				parent: #debugging;
+				default: true;
```
StevenCostiou: Could it be possible (technically) to directly put a color?
Either there are no color selected, or a custom color and then we can choose our color?
Maybe not easily possible...

StevenCostiou (changes requested):
Thanks Cyril!
A few comments, but seems good otherwise.

src/NewTools-Debugger/StDebugger.class.st
jecisc: Done!

src/NewTools-Debugger/StDebugger.class.st
jecisc: Good catch thanks :)

src/NewTools-Debugger/StDebugger.class.st
jecisc: IMO, since the color used comes from the theme, we should not override the theme. 

But one thing that would be good, would be to implement a 'configurable theme' and allow to customize it in the settings so that people can change just one color without having to implement a theme. But I don't think I'll have the tome to implement that in the near future :( Is that ok?

StevenCostiou: No worries, it is not a priority right now.

I have though a different perspective on customizing.
Why would we block users to override the theme?

For me it is really is killing my experience when I just want to put, for example, a window in red because that window is important for me, but the system prevents me from doing that because "I should not override the system".

I agree that a good solution would be to build a powerful customizable theme where we can change things easily, but since it is not going to happen anytime soon, why not allowing that kind of small customization?

Regarding the debugger, I would like to keep that discussion open, but that will be for another time ;)

StevenCostiou (approved):

Ducasse:
Cool. I just mentioned to Cyril that we should pay attention with too pale (or not even intense colors) because of color blind people. There are software that simulate what the color blind people can see and this is nice to try. We did it for the mooc template.

PR #523, opened 2023-05-12
Cleanup-AbstractTool 

MarcusDenker:
Rewrite all users of AbstractTool, move some feature to SystemNavigation (which is a horrible thing, too).

But this way we have one strange things removed (AbstractTool)

- rewrite users of lookupVar:declare:

### DIFF
--- a/src/NewTools-ChangeSorter/ChangeSorterModel.class.st
+++ b/src/NewTools-ChangeSorter/ChangeSorterModel.class.st
@@ -3,7 +3,7 @@ A ChangeSorterModel is a model used by Change Sorter UIs for computation
 "
 Class {
 	#name : #ChangeSorterModel,
-	#superclass : #AbstractTool,
+	#superclass : #Model,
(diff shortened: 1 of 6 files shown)


Ducasse:
I read the changes and it is unclear to me why self model is worse than SystemNavigation new. 
I could not understand why AbstractTool is bad either. 
Now we can proceed and integrate these changes.

MarcusDenker:
the model then was just forwarding to SystemNavigation for these methods

AbstractTool is just used by old tools.. e.g. the inspector or debugger do no use it. Just ChangeSorter (it's model was a subclass), MessageList (which was just doing AbstractTool new) and the too for Manifest (it's model was a subclass of AbstractTool).

We need something, but just *one* abstraction. This step removes AbstractTool, as the code was strange and it was mostly just forwarding to SystemNavigation. We then need to continue, e.g SystemNavigation with all the UI related methods is very strange.

PR #524, opened 2023-05-15
Do not listen to method recategorization in MessageBrowser

jecisc:
Remove announcement listening doing nothing


jecisc:
The error is not related to my change since I didn't add any new class in this PR

PR #530, opened 2023-05-20
Add source folder of Sindarin

jecisc:
In order to check if the error we get from the Pharo repository, let's try to add the "src" folder of sindarin to the URI


jecisc:
The tests are passing and this might fix the Pharo build so I'll merge it. Pharo build is broken currently so it cannot make things worse.

adri09070:
Ah, that was the problem, it's good to know !

I thought that Metacello would find the source folder thanks to the .project file

jecisc:
I’ll never sure when things will or will not find the source folder so I always specify it just in case :/

PR #537, opened 2023-05-30
StDebuggerTestFailureContextPredicate #printDescription can use #tempNamed:

MarcusDenker:
printDescription is looking up the var and then it reads it, we can use #tempNamed: to do that in one step


StevenCostiou (approved):
The failing test seems unrelated to the changes.

PR #539, opened 2023-05-31
Small compiler API useage fix in StDebuggerActionModel

MarcusDenker:
in StDebuggerActionModel>>#returnValueFromExpression:fromContext: wedo not need to set the context and the receiver


StevenCostiou (approved):
Failing tests seem unrelated

PR #540, opened 2023-05-31
Cleanup-debuggerColor

MarcusDenker:
debuggerColor is never used, it references SHRBTextStyler and one method even sends a non existing selector #debuggerColors


StevenCostiou (approved):
This was old code to change the color of the displayed node in the debugger.
This is not used anymore.

Failing tests seem unrelated.

PR #545, opened 2023-06-08
Use better protocol API

jecisc:
CompiledMethod>>#protocol behavior will change and a more explicit API has been added in Pharo to replace the current users that cannot deal with a real protocol. This comit updates the API to the new one using #protocolName or #isClassified


jecisc:
Test failures are not related. They are also present on other PRs.

MarcusDenker (approved):

PR #546, opened 2023-06-08
Make initial extent of StSpotter and StInspector take the display scale factor into account

Rinzwind:
This pull request makes the initial extent of StSpotter and StInspector take the display scale factor into account.


MarcusDenker:
as we merged the same changer on the pharo main code base, I propose to merge this, too

MarcusDenker (approved):

PR #553, opened 2023-06-23
Fix spotter advance search

jecisc:
Now this is the UI that is responsible for opening an UI.


MarcusDenker:
already merged as https://github.com/pharo-spec/NewTools/pull/557

PR #557, opened 2023-06-30
 Fixing opening of the Finder on the Spotter

jordanmontt:
Calling FinderUI instead of Finder since the method Finder>>open was moved.  Now the correct way is to call FinderUI>>open.

Pair programming with @uNouss for the sprint


Ducasse:
Tx!

MarcusDenker (approved):

PR #571, opened 2023-08-01
Use new API of ClassFactoryForTestCase

jecisc:
ClassFactoryForTestCase duplicates a lot of the old class creation API that is not extensioble and huge. A new way of creating classes through it is been implemented. Here is a PR with changes about this.

PR depends on this PR: https://github.com/pharo-project/pharo/pull/14398


jecisc:
The one failing test is not related and I did a PR to Pharo to fix it (it is probably not triggered in Pharo because of the order of tests)

Ducasse:
So cyril what should I do?

jecisc:
It’s mergeable now :)

MarcusDenker (approved):

PR #575, opened 2023-08-04
452 in the debugger create subclass responsibility button does not offers the possibility to create the responsibility in subclasses of the receiver but in the exception class subclass responsibility but only after performing steps 

### LINKED ISSUE #452: In the debugger, create subclass responsibility button does not offers the possibility to create the responsibility in subclasses of the receiver but in the exception class `SubclassResponsibility`, but only after performing steps
**To reproduce the bug:**

Class hierarchy:

image

image

In a playground:
```Smalltalk
obj := B new.
obj tata
``` 
and right-click > debug it on `obj tata`.

Perform STEPINTO > STEPOVER so that the debugger offers you the possibility to create the subclass responsibility.

Then click on the create button and the debugger will ask you to create the responsibility in `SubclassResponsibility class` or `SelectorException class` instead of `A` or `B`:

image

Note that the bug only appears when performing a step to get to the point where the subclass responsibility must be defined. If you perform a do it or proceed, the debugger will ask you correctly to create the subclass responsibility in `A` or `B`.

adri09070:
Fixes #452


jecisc:
There is a conflict

adri09070:
Failing test looks unrelated

jecisc:
Are you working in Pharo 11? Because I see that the Tonel format that was moved to the V3 version got back to the V1 version here

jecisc (approved):

PR #585, opened 2023-09-19
Creates Scopes Editor

carolahp:
## Summary
This PR implements a presenter for creating and editing scopes.
Scopes are used to define parts of the code where refactorings will be applied.

## How to test
Open the scope editor from the Pharo menu 
image

While testing consider the following:
- Calypso "Add to scope" commands in contextual menus are still not integrated to the scope editor, and will be replaced by new commands in the Pharo PR mentioned below

## Pending features
- Browse a scope on Calypso
- Integrate with Pharo 12, displaying the new scopes in the presenters related to refactorings (I will create a PR on Pharo after this PR is merged)


estebanlm:
is this actually a browser?
as far as I understand, is a tool to select a scope, not what we call a browser.

Ducasse:
We can call it a Scope Editor

estebanlm:
yep, that fits better :)

estebanlm:
yeah, but we need also a package rename ;)

estebanlm:
hi @carolahp 
can you please rename the package `NewTools-Scopes-Browser` to `NewTools-Scopes-Editor`?
the more time you take for doing it, more history we will lose in inceberg :(

carolahp:
> hi @carolahp can you please rename the package `NewTools-Scopes-Browser` to `NewTools-Scopes-Editor`? the more time you take for doing it, more history we will lose in inceberg :(

hi @estebanlm , it's done, I didn't realize we were losing history

Ducasse:
I will merge it so that we can try it.

estebanlm:
why things cannot be tested by just load them, instead merging them into the trunk?

Ducasse:
My point is not me just but all the people using the scope editor.

PR #595, opened 2023-10-04
fix #testStackTablePackagesLabels: use correct DoIt selector

MarcusDenker:
testStackTablePackagesLabels uses the wrong selector for the DoIt (thus #isDoIt would return false)


jecisc (approved):

jecisc:
Oh! It has been long since I didn't see a green CI on new tools :) This is nice, our recent fixes had a good impact

PR #601, opened 2023-10-13
SpChangeSorterPresenter shouldnt use ClassRecategorized

jecisc:
It should use ClassRepackaged instead


jecisc:
@MarcusDenker There is an easy review to do here

MarcusDenker (approved):

PR #629, opened 2023-11-27
[Proposal] Put debugger stack on the left

guillep:
imagen

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -765,31 +765,22 @@ StDebugger >> initializeStack [
 		activateOnDoubleClick;
 		whenActivatedDo: [ :selection | self doBrowseClass ].
 	stackTable
-		addColumn: ((SpImageTableColumn title: '' evaluated: [ :context | self stackIconForContext: context ])
+		addColumn: ((SpImageTableColumn
+				  title: ''
+				  evaluated: [ :context | self stackIconForContext: context ])
 				 width: 8;
 				 sortFunction: nil;
 				 yourself);
-		addColumn: ((SpStringTableColumn title: 'Class' evaluated: [ :context | self printReceiverClassInContext: context ])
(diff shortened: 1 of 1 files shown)


estebanlm:
not against :)

but I am working on the message browser, and I added a "flip" button to allow the user to switch as they please.
I guess it could be the same for the debugger? 
 
EDIT: submitted before finish, fixed :P

StevenCostiou:
Following this proposition, I planned to propose a dynamic switch between layouts in the debugger.
Not exactly a flip, more there is a list of layouts some provided by default and if you want you can add yours, and then you select the one that you want in the debugger (from a menu somewhere) and it becomes the new default in your image.

I wanted to do that last friday because it is really trivial to do, but I got sick with kids...

I'll try it tomorrow if I don't have kids on my hands :p

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -1147,7 +1138,7 @@ StDebugger >> stack [
 { #category : 'specs' }
 StDebugger >> stackAndCodeLayout [
 
-	^ SpPanedLayout newTopToBottom
+	^ SpPanedLayout newLeftToRight
 		positionOfSlider: 30 percent;
 		add: self stackLayout;
 		add: self codeLayout;
```
StevenCostiou, line 1144: This is actually the only change right?
All the other diffs are just code reformating? (hard to see)

StevenCostiou (approved):

src/NewTools-Debugger/StDebugger.class.st
guillep: I changed also the columns.
Having too many columns is good for a vertical layout, where the method list can be very large.
But for the horizontal layout the list should be narrower, so having too many columns is not very good there...
Tradeoffs :)

StevenCostiou:
We can close it now, since another PR implementing this change has been merged.

PR #635, opened 2023-12-05
Editable inspector values

### LINKED ISSUE #501: [Improvement] Can edit inspect object vars directly in an inspector
This can be cool to directly edit vars into an inspector, for example in this class :

image

I would like to edit the "title" of my object by double click on the field, typing the value and press enter.

hernanmd:
### Issue
Currently in Pharo 12, the inspector doesn't allow direct editing of object variables, requiring additional steps for modification.

### Proposed Solution
I propose enhancing the inspector to support direct editing of object variables, streamlining the debugging process and improving developer efficiency.

### Implementation
- Introduce inline editing for object variables.
- Ensure proper validation and handling of edited values.
- Maintain compatibility with existing inspector functionality.

This change aims to enhance the developer experience and reduce the time spent on debugging tasks.

Closes #501

### DIFF
--- a/src/NewTools-Inspector/StRawInspectionPresenter.class.st
+++ b/src/NewTools-Inspector/StRawInspectionPresenter.class.st
@@ -178,7 +178,9 @@ StRawInspectionPresenter >> valuesColumn [
 		title: 'Value';
 		evaluated: #stringValue;
 		beEditable;
-		onAcceptEdition: [ :node :value | self write: node value: value ];
+		onAcceptEdition: [ :node :value | 
+			self write: node value: value.
+			self window title: self window presenter windowTitle ];
 		sortFunction: #stringValue ascending;
 		yourself
 ]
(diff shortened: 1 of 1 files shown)


src/NewTools-Inspector/StRawInspectionPresenter.class.st
```diff
@@ -200,3 +211,13 @@ StRawInspectionPresenter >> variablesColumn [
...
 		yourself
 ]
+
+{ #category : 'writing' }
+StRawInspectionPresenter >> write: aStInspectorSlotNode value: aString [ 
+	"Update the value of aStInspectorSlotNode with new value aString.
+	Convert aString to the 'native' object parsing it as number or using parseLiterals (for booleans or strings)"
+	
+	aStInspectorSlotNode slot 
+		write: (NumberParser parse: aString onError: [ aString parseLiterals first ])
```
MarcusDenker, line 221: It would (later) be interesting to see if we can support evaluating an expression here... it should not be that much more dangerous

Issue #615, opened 2023-11-21
testUpdateExtensionSubscription is often failing on Pharo12


jecisc:
This test was previously failing from time to time on Pharo's CI but recently it is almost always failing on Pharo 12 CI but it's almost always on OSX. 

This is probably related to something random but maybe it can be used to find the origin? (For example doing a PR with logs in the test to find the origin?)

Example of stack

```
osx-64 / Tests-osx-64 / testUpdateExtensionSubscription – MacOSX64.NewTools.Debugger.Tests.StDebuggerTest8s
Error
Got 1 instead of 5.
Stacktrace
TestFailure
Got 1 instead of 5.
StDebuggerTest(TestAsserter)>>assert:description:resumable:
StDebuggerTest(TestAsserter)>>assert:description:
StDebuggerTest(TestAsserter)>>assert:equals:
StDebuggerTest>>testUpdateExtensionSubscription
StDebuggerTest(TestCase)>>performTest
```

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-15488/1/tests

Can maybe interest @StevenCostiou or @adri09070

adri09070:
How can I put logs in Pharo to see something on the CI ?

jecisc:
If you use a #traceCr it will show in the logs

adri09070:
Does the error still occur?

I think it happens because the system announcer subscriptions regsistry randomly removes its weak subscriptions that can be removed. 

I don't understand how it is possible that some debuggers are still registered to the system announcer after they have been garbage collected though

PR #638, opened 2023-12-06
Prevent font chooser open error listing

### LINKED ISSUE #637: Font family listing defaults to nil and prevents font chooser to open
Pharo has a limited set of fonts available by default which can be listed with:

```smalltalk
LogicalFontManager current allFamilies

 "'an Array(#FreeTypeFontFamily ''.Lucida Grande UI'' #TextStyleAsFontFamily #''Bitmap DejaVu Sans'' #TextStyleAsFontFamily #''Bitmap Source Code Pro'' #TextStyleAsFontFamily #''Bitmap Source Sans Pro'' #FreeTypeFontFamily ''Lucida Grande'' #FreeTypeFontFamily ''Source Code Pro'' #FreeTypeFontFamily ''Source Sans Pro'')'"
```

unless you evaluate:

```smalltalk
FreeTypeFontProvider current updateFromSystem.
```

Thus if you request a font chooser window with a non-default family (which is installed in the system) or a family not updated from system, the initialFamily is set to nil, preventing to get an initialStyle:

```smalltalk
| fontChooser presenter |

fontChooser := FontChooser newWithAllFamilies
	title: 'Title';
	font: (LogicalFont
		familyName: 'Arial'
		pointSize: 11
		stretchValue: 5
		weightValue: 300
		slantValue: 0);
	onAcceptDo: [ : e | self halt ].
presenter := StFontChooserPresenter on: fontChooser.
presenter open.
```

Screenshot 2023-12-06 at 15 35 04

hernanmd:
### Issue
Opening font chooser with a non-existent font family raises an exception.

### Solution
Try updating Font Chooser with the selected font; if not found, inform and select the first available font family and style.

Add tests

Fixes #637.

### DIFF
--- a/src/NewTools-FontChooser/StFontChooserPresenter.class.st
+++ b/src/NewTools-FontChooser/StFontChooserPresenter.class.st
@@ -357,7 +357,7 @@ StFontChooserPresenter >> updateWithFont: aFont [
 	initialFamily := fontFamilyList items
 		                 detect: [ :e | e familyName = aFont familyName ]
 		                 ifNone: [ 
-									self inform: ('Requested font {1} not found' format: { aFont familyName }).
+									('Requested font {1} not found' format: { aFont familyName }) traceCr.
 									fontFamilyList items anyOne ].
 
 	fontFamilyList selectItem: initialFamily.
--- a/src/NewTools-FontChooser/StFontChooserPresenter.class.st
+++ b/src/NewTools-FontChooser/StFontChooserPresenter.class.st
@@ -347,13 +347,18 @@ StFontChooserPresenter >> updatePreview [
 
 { #category : 'initialization' }
 StFontChooserPresenter >> updateWithFont: aFont [
+	"Try to update the receiver's listing with the selected aFont. 
+	If we do not find a font family matching aFont (for example not yet loaded fonts), inform and then select the first family in the system.
+	If we do not find a font style, select the first one available as initial style"
 
 	| initialFamily initialStyle |
 	fontFamilyList items: self model fontList.
 
 	initialFamily := fontFamilyList items
 		                 detect: [ :e | e familyName = aFont familyName ]
-		                 ifNone: [ nil ].
+		                 ifNone: [ 
+									self inform: ('Requested font {1} not found' format: { aFont familyName }).
+									fontFamilyList items anyOne ].
 
 	fontFamilyList selectItem: initialFamily.
 
@@ -362,7 +367,7 @@ StFontChooserPresenter >> updateWithFont: aFont [
 			                style weightValue = aFont weightValue and: [ 
 				                style stretchValue = aFont stretchValue and: [ 
 					                style slantValue = aFont slantValue ] ] ]
-		                ifNone: [ nil ].
+		                ifNone: [ self selectedFontFamily members anyOne ].
 
 	fontStyleList selectItem: initialStyle.
 
--- a/src/NewTools-FontChooser/StFontChooserPresenterTest.class.st
+++ b/src/NewTools-FontChooser/StFontChooserPresenterTest.class.st
@@ -38,6 +38,15 @@ StFontChooserPresenterTest >> newDefaultFamilyFontChooserFont: aLogicalFont [
 		yourself.	
 ]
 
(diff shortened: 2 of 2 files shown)


hernanmd:
The [failing test testUpdateExtensionSubscription is unrelated](#615) to the fix

src/NewTools-FontChooser/StFontChooserPresenter.class.st
```diff
@@ -347,13 +347,18 @@ StFontChooserPresenter >> updatePreview [
 
 { #category : 'initialization' }
 StFontChooserPresenter >> updateWithFont: aFont [
+	"Try to update the receiver's listing with the selected aFont. 
+	If we do not find a font family matching aFont (for example not yet loaded fonts), inform and then select the first family in the system.
+	If we do not find a font style, select the first one available as initial style"
```
demarey, line 353: I'm not sure I would select the first one. 
Why not just abort the current update of font?

src/NewTools-FontChooser/StFontChooserPresenterTest.class.st
```diff
@@ -8,16 +8,53 @@ Class {
...
+		yourself.	
+]
+
+{ #category : 'tests' }
+StFontChooserPresenterTest >> openPresenterShouldntRaiseError [
+
+	self shouldnt: [ 
+			| window | 
+			window := presenter open.
+			window close ] raise: Error
```
demarey, line 47: The goal of a test is to capture an error if any.
Then `shouldnt: raise: Error` is useless but the test is also useless. 
The method should just be named openPresenter.

src/NewTools-FontChooser/StFontChooserPresenter.class.st
hernanmd: Christoph could you reproduce the bug in your image? (note that it requires you to select a non default font). It is not about canceling to open the lookup but to prevent an error when you have already a configured font.

src/NewTools-FontChooser/StFontChooserPresenterTest.class.st
hernanmd: I don't understand. The presenter configured in the setUp fails to open if the fix in the PR is not applied. As stated in the issue:

```smalltalk
| fontChooser presenter |

fontChooser := FontChooser newWithAllFamilies
	title: 'Title';
	font: (LogicalFont
		familyName: 'Arial'
		pointSize: 11
		stretchValue: 5
		weightValue: 300
		slantValue: 0);
	onAcceptDo: [ : e | self halt ].
presenter := StFontChooserPresenter on: fontChooser.
presenter open.
```

See the following screenshot:
Screenshot 2023-12-11 at 16 46 45

The method is named openPresenterShouldntRaiseError since it is common code called from multiple tests:

```
StFontChooserPresenterTest>>#testOpenAllOnNonExistantFontFamily StFontChooserPresenterTest>>#testOpenDefaultOnUnloadedFontFamily StFontChooserPresenterTest>>#testOpenDefaultOnNonExistantFontFamily StFontChooserPresenterTest>>#testOpenAllOnUnloadedFontFamily
```

src/NewTools-FontChooser/StFontChooserPresenter.class.st
demarey: I just reproduced the problem.
I do not know where this presenter is used but If I understand correctly, this presenter can only works if a font is selected, right?
In this case, the fix is ok.

src/NewTools-FontChooser/StFontChooserPresenterTest.class.st
demarey: What I mean is that a method test should never include:
```smalltalk
self shouldnt: [ ] raise: Error
```
It is the purpose of the test framework to capture all Errors and to make your test fail if any.
It would only make sense if you wanted to check for a specific error that should not happen.
You should rename `StFontChooserPresenterTest>>#openPresenterShouldntRaiseError` to `StFontChooserPresenterTest>>#openPresenter` with:
```smalltalk
openPresenter

	| window | 
	window := presenter open.
	window close 
```

PR #643, opened 2023-12-08
Test debugger does not log exceptions twice when opened

hernanmd:
Add a test for [PR in Pharo](https://github.com/pharo-project/pharo/pull/15746), to check that when the log debugger stack to file setting is false, a Pharo log file is not created when requesting a debugger.


src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -506,6 +506,31 @@ StDebuggerTest >> testDebuggerApplication [
...
+	debugger
+		application: debugger class currentApplication;
+		initialize;
+		open.
+
+	self 
+		assert: logFileRef exists not
+		description: 'It tests that when the log debugger stack to file setting is false, a Pharo log file is not created when requesting a debugger'.
+
+	session class logDebuggerStackToFile: debuggingStackToFile
```
adri09070: If the test fails, it will not set back the `debuggingStackToFile` setting. Maybe, the assignment should be done in an `ensure:` block or maybe it should be set in the `tearDown:`

adri09070 (changes requested):
The test is really nice to have, thanks!

Just one remark though:

If the test fails, it will not set back the `debuggingStackToFile` setting. Maybe, the assignment should be done in an `ensure:` block or maybe it should be set in the `tearDown:`

adri09070 (approved):

PR #644, opened 2023-12-12
Fixed SubscriptionRegistry inspection freezing image

### LINKED ISSUE #15005: Cannot inspect SubscriptionRegistry of SystemAnnouncer
In Pharo 12 if I inspect the unique instance of the system announcer and click in the raw view on the registry then my image freeze and I cannot do anything

JanBliznicenko:
Fixes https://github.com/pharo-project/pharo/issues/15005

There was a table of subscriptions in both `Announcer` and `SubscriptionRegistry`, showing same items, yet each implemented separately and differently, only `SubscriptionRegistry` causing the issue.

This PR replaces the implementation in `SubscriptionRegistry` by the one from `Announcer`, as well as making the one in `Announcer` just call the one from `SubscriptionRegistry`, thus removing duplicated code.


jecisc (approved):
Nice thanks

PR #645, opened 2023-12-14
Remove change sorter

jecisc:
This code is not loaded and will never be loaded again. 

I also removed the last dependency of NewTools over ChangeSet (If I didn't miss any)


Gabriel-Darbord (approved):

Ducasse:
Youpiiiiiiii

PR #661, opened 2024-01-21
Update New Finder log errors to Transcript setting

hernanmd:
This PR changes the default in logErrorsToTranscript setting, so we do not pollute output by default.
Also execute the Example tests without logging to transcript so the CI isn't polluted.


Ducasse:
Hi hernan 
why do we expect to have messageNotUnderstand in the finder (no need to reply now you can explain it to me next time we meet).

PR #662, opened 2024-01-21
Fix Undeclared reference in StClyMessageListPresenter

hernanmd:
Fixes testUndeclared reported by ReleaseTest during CI 
RBRemoveMethodDriver -> ReRemoveMethodDriver in StClyMessageListPresenter


Ducasse:
Tx Hernan!
I will integrate also the large renaming I did on Refactorings.

PR #667, opened 2024-01-23
The debugger now delegate UI decisions to the backend provided by its application

StevenCostiou:
It requires the spec gtk backend to reimplement spawnNewDebugSessionForSpecBackend: and returns its own debug session.
The spec gtk debug session must implement at least deferDebuggerOpeningToBackend: that can just do aStDebugger openWithFullView.

Requires https://github.com/pharo-project/pharo/pull/16015


StevenCostiou:
This should now be in conflict with latest merges in the debugger.
I will take care of that but not immediately (there is a PR waiting list...)

Ducasse:
@StevenCostiou I moved the PR to Pharo 13 and will see if I can find a way to relaunch the tests

StevenCostiou:
Conflict solved.
Waiting for tests to run.

PR #668, opened 2024-01-26
fixes the "bytes" view of the inspector for CompiledCode

MarcusDenker:
fix for https://github.com/pharo-project/pharo/issues/16018


estebanlm:
this was fixed already here: https://github.com/pharo-spec/NewTools/pull/670

PR #669, opened 2024-01-28
Changed StFinderTest superclass to SpBaseTest

hernanmd:
As reported in https://github.com/pharo-project/pharo/issues/16041 to avoid problems during concurrent accessing of morphs from within the CL handler and Morphic UI processes. Fix suggested by @Rinzwind


hernanmd:
Errors are unrelated.
(some of them addressed in this PR: https://github.com/pharo-vcs/iceberg/pull/1797)

PR #670, opened 2024-01-28
Fix bytes panel is broken on compiled methods

hernanmd:
As reported in https://github.com/pharo-project/pharo/issues/16018
Applied the fix proposed by @MarcusDenker


Ducasse:
With marcus we did one where we introduced a hook but this is ok too.

PR #681, opened 2024-02-02
New settings browser

hernanmd:
This PR includes a new implementation of the Settings Browser based in Spec 2, which does not have Morphic dependencies in its core. To launch it you can:

```smalltalk
StSettingsApplication open.
```

or 

Library -> New settings browser

## Notes

- Implementation is not complete. It needs more usage and testing.
- Update Baseline to include the new setting browser and color picker dependency.
- Add setting browser first set of tests.
- Many configurations rely on installing method extensions on different classes. These methods will later be repackaged. The classes in which method extensions were installed are:
  - UITheme
  - ThemeIcons
  - NECPreferences
  - SHStyleElement
  - LogicalFont


Ducasse:
Super nice news!!!!!
I will check because I do not really know how to load it in my image to give a try. So we can push it and iterate?

hernanmd:
> Super nice news!!!!! I will check because I do not really know how to load it in my image to give a try. So we can push it and iterate?

Yes please, this is mostly to iterate from smaller commits.

Ducasse:
Broken test is unrelated. So I merge so that we move on.

PR #683, opened 2024-02-08
Compression UI utilities

hernanmd:
Re-take of [PR](https://github.com/pharo-project/pharo/pull/16106)'s causing dependency problems in the build.
In this PR, a new package is proposed: `NewTools-Compression-Utils`.

The new package includes a couple of minimal method to add open directory dialog to Compression classes.
It also removes the UIManager dependency from Compression.

These changes were the same as the previous PR's, but does instead of using the SpApplication hook, it uses directly the new FileBrowser API.

### DIFF
--- a/src/BaselineOfNewTools/BaselineOfNewTools.class.st
+++ b/src/BaselineOfNewTools/BaselineOfNewTools.class.st
@@ -138,20 +138,20 @@ BaselineOfNewTools >> baseline: spec [
 						'Core' 
 						'NewTools-FlagBrowser' 
 						'NewTools-FlagBrowser-Tests' );
-			group: #development with: #( 'default' 
+			group: 'development' with: #( 'default' 
 						'NewTools-DebuggerSelector'
 				   		'NewTools-DebuggerSelector-Tests' );
-			group: #FileBrowser with: #( 
+			group: 'FileBrowser' with: #( 
 						'NewTools-FileBrowser' 
 						'NewTools-FileBrowser-Tests' );
-      			group: #RewriterTools with: #(
+      			group: 'RewriterTools' with: #(
 				    		'NewTools-RewriterTools-Backend'
 				    		'NewTools-RewriterTools'
 				    		'NewTools-RewriterTools-Backend-Tests'
 				    		'NewTools-RewriterTools-Tests' );
 			group: 'Profiler' with: #( 'NewTools-ProfilerUI' );
 			"ScopesEditor"
-			group: #ScopesEditor with: #(
+			group: 'ScopesEditor' with: #(
 						'NewTools-Scopes' 
 						'NewTools-Scopes-Editor' 
 						'NewTools-Scopes-Resources-A-Tests' 
@@ -159,11 +159,11 @@ BaselineOfNewTools >> baseline: spec [
 						'NewTools-Scopes-Resources-C-Tests' 
 						'NewTools-Scopes-Tests');
 			"Finder"
-			group: #Finder with: #(
+			group: 'Finder' with: #(
 						'NewTools-Finder'
 						'NewTools-Finder-Tests');
 			"Settings Browser"
-			group: #SettingsBrowser with: #(
+			group: 'SettingsBrowser' with: #(
 						'NewTools-SettingsBrowser'
 						'NewTools-SettingsBrowser-Tests');
 			"Compression Utilities"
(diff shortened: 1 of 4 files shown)


hernanmd:
Failing test `testTransformingDeprecation` is not related.

guillep (commented):
Looks like this approach will work better, but I'll wait for @estebanlm here :)
This adds a dependency from NewTools to Compression, which is understandable to me, and could be further refactored later. But maybe Esteban is thinking on a different package/project/repo organization.

Ducasse:
This looks good to me because the UI and core of compression have been separated.

Ducasse:
This is strange that we still get this traits fileout bug

hernanmd:
> This is strange that we still get this traits fileout bug

Yes, I have to unselect definitions from Iceberg before commiting but some can slip away :(

guillep:
Seems my PR solving that was not yet integrated :)

https://github.com/pharo-project/pharo/pull/15978

There is an issue and I have to check it.
I'll try to do it quickly.

estebanlm:
I can't be sure the remaining error is because of the PR or inherited. 
Please review :)

guillep:
> I can't be sure the remaining error is because of the PR or inherited.

I re-run the job to see if we can discard the second option.

>  Please review :)

Could you also give us your opinion on the issue and the package structure here?
Besides the broken test is the rest ok?

estebanlm:
yeah, is ok. I would not use a symbol to declare a group because it is not how the other groups are defined, but that is just a detail.

src/BaselineOfNewTools/BaselineOfNewTools.class.st
```diff
@@ -164,6 +166,9 @@ BaselineOfNewTools >> baseline: spec [
 			group: #SettingsBrowser with: #(
 						'NewTools-SettingsBrowser'
 						'NewTools-SettingsBrowser-Tests');
+			"Compression Utilities"
+			group: #CompressionUtilities with: #(
```
guillep:
```suggestion
			group: 'CompressionUtilities' with: #(
```

guillep (approved):
Ok! Now let's wait for the CI.

src/BaselineOfNewTools/BaselineOfNewTools.class.st
estebanlm: I see there are several groups that are using symbols now (it was  not like that before). Since you are there, can you align them too?

guillep: I did it from the github PR. You're here too :)

estebanlm (approved):
those errors do not seem related

src/NewTools-Compression-Utils/InflateStream.extension.st
```diff
@@ -0,0 +1,26 @@
+Extension { #name : 'InflateStream' }
+
+{ #category : '*NewTools-Compression-Utils' }
```
guillep, line 3: Now that this method was moved here, is there a "sister" PR in the Pharo repository removing the extension from there?

src/NewTools-Compression-Utils/InflateStream.extension.st
```diff
@@ -0,0 +1,26 @@
...
+	SpTextPresenter new
+		text: contentsString;
+		open;
+		withWindowDo: [ : w | 
+			w 
+				title: titleString;
+				extent: 600 @ 800 ]
+]
+
+{ #category : '*NewTools-Compression-Utils' }
```
guillep, line 16: Same

src/NewTools-Compression-Utils/ZipArchive.extension.st
```diff
@@ -0,0 +1,18 @@
+Extension { #name : 'ZipArchive' }
+
+{ #category : '*NewTools-Compression-Utils' }
```
guillep, line 3: Same

guillep (approved):
Hi @hernanmd

I think we are good to go.
Can you check my comment? The extensions that were moved here should be removed from Pharo (otherwise they will be duplicated and iceberg will confuse people :) )

guillep:
Thanks @estebanlm ! I'll merge!

src/NewTools-Compression-Utils/InflateStream.extension.st
hernanmd: Ok, if there's no rush I will check it later today since now I'm about to start a meeting

guillep: Sure, it's not for yesterday :)

PR #684, opened 2024-02-12
Add missing methods in Finder

hernanmd:
This is a small PR for the new Finder to add a missing method preventing to browse the selected item in the results list.

- Add missing #isClassResult method.
- Add key bindings to result tree.


hernanmd:
Tx!

PR #689, opened 2024-02-14
Improve File Browser

jecisc:
This improves multiple things:

- Add tooltip to the icons
- Fix a bug where we could only edit the path once then the button became useless
- Change the settings icon to use the same as the Pharo menu one
- Do not make the hiding of hidden files a filter but a setting impacting all filters. There is now an entry in the settings and also a command to toggle the value.


hernanmd (approved):
Changes are ok

PR #694, opened 2024-02-14
Deprecate asFileName and re-write senders

hernanmd:
As reported in [this pharo issue](https://github.com/pharo-project/pharo/issues/16123), this PR should be applied along with the [PR in pharo repository](https://github.com/pharo-project/pharo/pull/16169). It is a small PR to clean dependencies which could bring problems when removing old FileList.


Ducasse:
I have a question why the code is not "as simple as"
`FileReference new withPath: ....`

hernanmd:
> I have a question why the code is not "as simple as" `FileReference new withPath: ....`

Apparently the file name re-"encoding" is just to replace characters such as $/ with $# before saving the file.
It is more a direct translation, but I never used `newWithPath:` so I'd need to check.

Ducasse:
What I means is that if String >> asFileName is doing something too complex it means that such complex behavior should be moved to fileReference so that 
- asFileName can be rewritten in a simpler way 
- users of asfilename can be rewritten simply too. 
So may be the logic of asFileName should be migrated to FileReference

guillep (approved):
Thanks @hernanmd ! I like the tests.

Reading your tests I wonder why is this `fuelOut` useful... But that's another story

jecisc:
I'm restarting the CI because I have the impression the error is not related

Ducasse:
Hi @hernanmd
Can you have a look because I would like to integrate it?

Ducasse:
@jecisc I wonder if your merging of packages did not impact this PR
can you have a look?

hernanmd:
> Hi @hernanmd Can you have a look because I would like to integrate it?

In #781

PR #697, opened 2024-02-15
Debug Point Browser integration

adri09070:
Needs https://github.com/pharo-project/pharo/pull/16177 to be merged first.

This PR integrates the debug points browser, a new tool to visualize and configure existing debug points. (Full details on the new debug point model are described in the PR above)

The tool can be opened via the `Debug > Debug Point Browser` world menu and looks like that: 

image

On the top-left, a table allows to visualize all debug points (breakpoints/watchpoints/etc.)
Each debug point has a name (more a tag than a name) that can be edited directly in this table.
The "enabled" state of the debug point is also displayed, as well as the scope of the debug point (class or object)

All behaviors of the selected can be configured via the right pane of debug point browser

When a debug point is selected, according to the target of the debug point, different information are displayed.
If the target is an AST node, the code of the concerned method is displayed (like in screenshot below), while highlightingthe corresponding code
If the target is an instance variable, all concerned methods are listed and selecting a method display displays the corresponding code and highlights the correct variable accesses:

image

It is possible to filter the displayed debug points by name, thanks to an input field at the top:

image

To set object-centric debug points, the left-most button in the inspector toolbar allows to do the same thing in the same way as current object-centric breakpoints:

image

Last but not least, it is also possible to change the scope of an existing debug point to an object, which was not possible before.
The third button in the inspector toolbar allows to do that and opens a modal to choose a debug point whose scope should be changed to the inspected object:

image

### DIFF
--- a/docs/debug-points-browser.md
+++ b/docs/debug-points-browser.md
@@ -1,0 +1,48 @@
+# Debug Points Browser
+
+The debug points browser is a new tool to visualize and configure existing debug points.
+Full details on the new debug points model are [described here](https://github.com/pharo-project/pharo/blob/Pharo12/doc/DebugPoints/debug-points.md)
+
+The tool can be opened via the `Debug > Debug Point Browser` world menu and looks like that: 
+
+![Opening Debug Point Browser via the World menu](https://github.com/pharo-spec/NewTools/assets/97704417/1cacd05d-f459-4a5c-bd9e-6587a89d6a95)
+
+On the top-left, a table allows to visualize all debug points (breakpoints/watchpoints/basic debug points/etc.)
+Each debug point has a name, which we could assimilate to a tag, that can be edited directly in this table.
+The "enabled" state of the debug point is also displayed via a checkbox, as well as the scope of the debug point (class or object).
+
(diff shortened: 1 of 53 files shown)


adri09070:
Maybe this failing test is related, I should investigate: 
`LinkInstallerTest>>#testSlotOrVarLinksRemainAfterMethodModificationForObject`

adri09070:
Needs https://github.com/pharo-project/pharo/pull/16268 to be merged first because it relies on the Calypso package that will be loaded after that PR is merged

adri09070:
To me, this is now  ready for review @StevenCostiou

estebanlm:
the prerequisites are merged. 
I am rerunning the tests.

estebanlm:
failing test is not related. Is @StevenCostiou approves, this can be merged

StevenCostiou (approved):

Ducasse:
Synced with steven -> ok

StevenCostiou:
There are a lot of changes, but only additions (the browser).
It should be merged and be experimented.
There will not be any impact if it is not used.

PR #701, opened 2024-02-16
Allow to sort some file browser columns

jecisc:
This change allows one to sort some of the File browser columns (name, size, date of creation...)


Ducasse:
sweeeeeet! Tx

PR #713, opened 2024-03-05
GT does not exist in Pharo so we can clean this test

hernanmd:
As reported in https://github.com/pharo-project/pharo/issues/16271


estebanlm:
no, is not :)
This test is not related to GT except for the name, which is old :P
We need to see why it started to fail now.

estebanlm:
I fixed the test

PR #716, opened 2024-03-12
Fix title display in File Presenter windows

hernanmd:
This PR enables to set the title of the file system presenters, as reported in https://github.com/pharo-project/pharo/pull/16084 by @demarey


hernanmd:
Failing test is unrelated

PR #720, opened 2024-03-18
Showing CMD character in OSX in Spotter

tesonep:
- The StPharoApplication should reset the backend when starting the image.
- The Spotter help should have the "shortcut" class to correct show the CMD character
Fix partially Pharo #14700


estebanlm (approved):
I really do not like that formatting. Otherwise is ok :)

PR #726, opened 2024-03-21
Fix code highlighting in debug points browser, for debug points set on AST node

### LINKED ISSUE #725: Debug Point Browser does not highlight corresponding code when a debug point is attached to a node
When you open the debug point browser, if you select a debug point set on an AST node in the list of debug points, the code presenter in the bottom-left won't highlight the code corresponding to the target node.

However, the highlight is done correctly for debug points set on variables

adri09070:
Fixes #725 

When you select a debug point in the debug point browser, it updates itself according to the target of the selected debug point.

In the case of an AST node, it updates the code presenter (code and  highlighting) and updates its layout to display this code presenter:

```Smalltalk
DebugPointNodeTarget >> updateDebugPointUIManager: aDebugPointUIManager [

	aDebugPointUIManager updateCode: self.
	aDebugPointUIManager switchToNodeTargetView
]

DebugPointCodePresenter >> updateCode: aDebugPointNodeTarget [
	"updates the code window with the code of the selected debug point"
        ... 
	self ... ;
		addTextSegmentDecoration: 
			(SpTextPresenterDecorator forHighlight
         		interval: (aDebugPointNodeTarget node start to: ((aDebugPointNodeTarget node stop) +1));
            yourself) .
]

DebugPointBrowserPresenter >> switchToNodeTargetView [

	self switchToTargetView: dpCode
] 

DebugPointBrowserPresenter >> switchToTargetView: aDebugPointTargetPresenter [
         "targetContainer is a SpPanedLayout"
	targetContainer replaceSecond: aDebugPointTargetPresenter
]
```

The method `#updateCode:` calls the method `#addTextSegmentDecoration:` that asks an adapter to add highlighting to the corresponding morphs.

However, for an obscure reason, `SpPanedLayout>>#replaceSecond:` called from `DebugPointBrowserPresenter >> switchToNodeTargetView` seems to replace the morphs corresponding to the presenter, and thus the highlight that was just added disappears.

To fix that, I just reversed the order of method calls in `DebugPointNodeTarget >> updateDebugPointUIManager:`.
First, we display the code presenter (which replaces the morphs corresponding to the code presenter) and then we update this code presenter (we add the highlight). In this order, the highlight remains and is correct


adri09070:
@StevenCostiou

jecisc (approved):
Seems good to me. Should I merge?

adri09070:
@StevenCostiou

StevenCostiou (approved):

PR #727, opened 2024-03-22
enable / disable syntax highlight in code presenter of critics browser.

### LINKED ISSUE #16290: CodeCritics displays rationale with method syntax hightlight
and it should be plain text.

demarey:
Fixes https://github.com/pharo-project/pharo/issues/16290


Ducasse:
I just removed the self flag: too

PR #733, opened 2024-04-03
Removing old object-centric breakpoint's integration to the inspector

adri09070:
All buttons and context menu entries to add object-centric breakpoints should be removed because we should use the new object-centric debug poins.

As a result, I removed:

- halt on read/write/state access in the raw inspector
- halt on read/write/state access in the object inspector
- halt on call/call once in the meta browser

@StevenCostiou


adri09070:
I don't think that the failing tests are related but I would request a confirmation from someone

estebanlm:
sounds like you're right :)
waiting for validation

estebanlm:
ping @StevenCostiou :)

StevenCostiou (approved):
The changes seem good, and have no impact.
Normally the removed actions are now available as debug points.

However the following failing tests are suspicious, as object-centric breakpoints depend on them:

Capture d’écran 2024-04-12 à 11 17 46

We should check that either they happen on other builds, or make sure that they are really disconnected.

adri09070:
> The changes seem good, and have no impact. Normally the removed actions are now available as debug points.
> 
> However the following failing tests are suspicious, as object-centric breakpoints depend on them:
> 
> Capture d’écran 2024-04-12 à 11 17 46
> 
> We should check that either they happen on other builds, or make sure that they are really disconnected.

I think this is weird too, but it indeed happens on other builds, for example this one from 2 weeks ago: https://github.com/pharo-spec/ScriptableDebugger/actions/runs/8479948371/job/23240927492

estebanlm:
uhm... I'll merge. But we will need to take care about those problems :(

PR #736, opened 2024-04-04
Fix issue while trying to display file setting items

hernanmd:
This PR fixes various problems with the setting browser:

- Add two setting wrappers for selection of directory and files in the new settings browser.
- Fix layout of settings pages to avoid unnecessary scroll bars.
- Better layout and colors.
- Add missing method to display file reference names in the setting browser.
- Remove duplicated `default:` in Playground setting. 
- Fix opening of Color Picker


hernanmd:
Same failing tests as in other previous PR (#733) so shouldn't be related

```
testCheckExistenceOfNonExistingFileDoesNotThrowException
testFlushLargeMessageOtherEndClosed
testSlotOrVarLinksRemainAfterMethodModificationForObject
```

PR #738, opened 2024-04-07
Update playground shortcut

jecisc:
This should be integrated with this: https://github.com/pharo-project/pharo/pull/16422


estebanlm:
the error is relevant :P

jecisc:
Yes because it needs the Pharo PR to be merged first

jecisc:
Now failing tests are unrelated :p

PR #739, opened 2024-04-09
New debugger extension model to display dynamic extensions

adri09070:
# New Dynamic Debugger Extension System

This PR introduces a new debugger extension model to display extensions dynamically. These "dynamic" extensions are displayed only if they are activated (that goes without saying) **and** only if we are in a specific context.

## Description of the New Debugger Extension System

The debugger extension determines if it can be displayed in a debugger via the method: 

```Smalltalk
TStDebuggerExtension>>#acceptsPredicate:  aStDebuggerContextPredicate

   ^ true "By default, a debugger extension is static"
```

This method uses the debugger's context predicate to determine if the extension should be displayed or not. **We could discuss: Should the method use a debugger as parameter instead of a context predicate?** The extension would have more parameters to decide if it should display itself or not, but is it necessary ?

This induces two big changes in the update of extensions:
- When an extension is (de)activated which sends an `StDebuggerExtensionToggleAnnouncement` to the debugger, the debugger now should consider the current context predicate,  because if it is refused by the toggled extension, the extension toggle should be ignored: it has no effect
- The context predicate can change after each action in the debugger, including steps (correct me if I'm wrong). So the extensions should be updated accordingly. Instead of just updating the instanciated presenters, we should check if there are:
  * extensions that were activated but that didn't accept the debugger context predicate until now, in which case it should be added now.
  * extensions that were activated and that did accept the debugger context predicate until now, in which case it should be removed from the extension tools.

This can be summed up by the two following boolean tables.

**Boolean table describing the action to take according to the current debugger context predicate when a debugger extension has been (de)activated, resulting in a `StDebuggerExtensionToggle` announcement:**

| `showInDebugger` state change | accepts predicate | action on extension |
| ---------------------------------| ------------------ | -------------------- |
| / | `false` | nothing |
| `false` -> `true` | `true` | **display** |
|  `true` -> `false` | `true` | **hide** | 

**Boolean table describing the action to take according to the current `showInDebugger` state of an extension, when  an action has been performed in the debugger, resulting in a potential change of status in the debugger context predicate:**

| `showInDebugger` state | accepts predicate change | action on extension |
| -------------------------| -------------------------- | -------------------- |
| `false` | / | nothing |
| `true` | `false` -> `false` | nothing |
| `true` | `true` -> `true` | nothing |
| `true` | `false` -> `true` | **display** |
| `true` | `true` -> `false` | **hide** |

## Introducing a New Dynamic Debugger Extension: Creating Method Extension

This debugger extension is displayed by default in the system settings:

image

This extension only appears if a DNU is raised and opens a debugger:

image

It has the exact same function as the "Create" button in the debugger toolbar, except that this extension allows the user to create the missing method while choosing the class or trait in the class hierarchy in which it should be installed, as well as the protocol under which it should be installed.
Whereas the "Create" button creates the missing class in the class of the receiver of the missing message, under a protocol chosen by the system.

With this extension, once you create the method by clicking the "Add method" button, the method has been created, the extension is not displayed anymore, and you can then proceed the execution (or debug it, of course):

image


adri09070:
@StevenCostiou

StevenCostiou:
The failing test seem unrelated?
I am reveiwing the changes now.

adri09070:
> The failing test seem unrelated? I am reveiwing the changes now.

The test fails in a fresh Pharo 13 image so it is unrelated

src/NewTools-Debugger-Exceptions/StDebuggerMNUExtensionTest.class.st
```diff
@@ -0,0 +1,154 @@
...
+{ #category : 'private - testing' }
+StDebuggerMNUExtensionTest >> extensionClassesOfDebugger [
+
+	^ debugger extensionTools collect: [ :extension | extension class ]
+]
+
+{ #category : 'accessing' }
+StDebuggerMNUExtensionTest >> extensionTested [
+
+	^ StDebuggerMNUExtension
```
StevenCostiou, line 27: Perhaps ask a MNU exception the class of the extension?

StevenCostiou (changes requested):
Good for merge.
Just one minor change that would be great ;)

src/NewTools-Debugger-Exceptions/StDebuggerMNUExtensionTest.class.st
StevenCostiou: Ok I take it back, I mixed up things.

PR #743, opened 2024-04-28
the tests use #category: from the package Deprecated12

pavel-krivanek:
required for https://github.com/pharo-project/pharo/pull/16538


jecisc:
For the CI to build we need P13 in smalltalk CI :(

jecisc (approved):

jecisc (changes requested):
I approved too fast. #package: on the shift class builder takes a string and not a package I think?

pavel-krivanek:
ouch, probably, fixed

PR #758, opened 2024-05-18
Remove deprecated classes and methods

jecisc:
Remove the deprecated classes and methods to start clean for P13


jecisc:
This still needs some work because some parts of Pharo and Roassal are using the deprecated classes

jecisc:
Should be fixed when we will have an image with https://github.com/pharo-project/pharo/pull/16770 and https://github.com/pharo-project/pharo/pull/16769

PR #762, opened 2024-05-29
[P12][P13] fix `#needRequestorScope` of `StDebuggerContextInteractionModel` so that its bindings can be used when evaluating code in the debugger

adri09070:
Currently, `StDebuggerContextInteractionModel >> #needRequestorScope` returns false, which prevents the debugger from using the bindings of its interaction model when evaluating code.

It should return `true`.

[For more details, see discussion here](https://github.com/pharo-spec/Spec/issues/1506)

This should be merged to Pharo13 AND Pharo12


adri09070:
unrelated failing test

MarcusDenker (approved):

PR #764, opened 2024-05-31
Updated text to new shortcut

jordanmontt:
Now the Playground's shortcut is `O + P` but the text was using O + W. This updates to the new Shortcut.

Before:
image

Now:
image


Ducasse:
Thanks seb.

PR #766, opened 2024-05-31
Update after selection in Color Picker

hernanmd:
Update each settings browser button color when a color is accepter in the Color Picker window.

Remove unnecessary initializePresenters in StSettingSectionPresenter.

### DIFF
--- a/src/NewTools-SettingsBrowser/Color.extension.st
+++ b/src/NewTools-SettingsBrowser/Color.extension.st
@@ -3,8 +3,19 @@ Extension { #name : 'Color' }
 { #category : '*NewTools-SettingsBrowser' }
 Color >> asSettingPresenter: aSettingDeclaration [ 
 
-	^ (StSettingSHStyleButtonColorPresenterItem on: aSettingDeclaration)
-		buttonAction: [ aSettingDeclaration getColor ];
-		whenColorChangedDo: [ self updateColor: aSettingDeclaration color ];
+	| settingPresenter colorValue |
+	
+	colorValue := aSettingDeclaration targetSelector 
+		ifNil: [ aSettingDeclaration realTarget perform: aSettingDeclaration name ]
+		ifNotNil: [ 
+			| settingsTarget |
+			settingsTarget := aSettingDeclaration realTarget perform: aSettingDeclaration targetSelector.
+			settingsTarget perform: aSettingDeclaration getSelector ].
+
+	settingPresenter := (SpPresenter new instantiate: StSettingSHStyleButtonColorPresenterItem on: aSettingDeclaration).
+	settingPresenter
+		buttonAction: [ aSettingDeclaration getColorFor: settingPresenter ];
+		buttonColor: colorValue;
 		yourself.
+	^ settingPresenter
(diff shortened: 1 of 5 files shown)


hernanmd:
Fixes #767

src/NewTools-SettingsBrowser/Color.extension.st
```diff
@@ -3,8 +3,19 @@ Extension { #name : 'Color' }
...
+	| settingPresenter colorValue |
+	
+	colorValue := aSettingDeclaration targetSelector 
+		ifNil: [ aSettingDeclaration realTarget perform: aSettingDeclaration name ]
+		ifNotNil: [ 
+			| settingsTarget |
+			settingsTarget := aSettingDeclaration realTarget perform: aSettingDeclaration targetSelector.
+			settingsTarget perform: aSettingDeclaration getSelector ].
+
+	settingPresenter := (SpPresenter new instantiate: StSettingSHStyleButtonColorPresenterItem on: aSettingDeclaration).
```
estebanlm, line 15: this is equivallent to just `StSettingSHStyleButtonColorPresenterItem on: aSettingDeclaration`, no need of the bureaucracy... and in any case is not recommended to do it like that. Which is the application that presenter belongs to?
(this still can work if you will after add that presenter to another that knows its app, or set the app, but *as is*, it doesn't) 
The recommended way of doing this is by passing a builder (see `SpPresenterBuilder`)

src/NewTools-SettingsBrowser/SettingDeclaration.extension.st
```diff
@@ -10,21 +10,23 @@ SettingDeclaration >> asSettingPresenter [
...
-SettingDeclaration >> getColor [
+SettingDeclaration >> getColorFor: aPresenter [
 
-	(SpPresenter new instantiate: SpColorPickerWindow on: self)
-		whenColorChangedDo: [ : c | 
+	| colorPickerPresenter |
+
+	colorPickerPresenter := aPresenter instantiate: SpColorPickerWindow on: self.
+	colorPickerPresenter
+		whenChangedDo: [ : c |
```
estebanlm, line 19: I prefer real names "aColor" or something like that.

estebanlm (changes requested):
please review those small issues (specially the first one).
notice that *it is possible* that you do not want a builder, in that case, please ensure in the other side (the one that will use the result of `asSettingPresenter`), there is an `owner` set to the resulting presenter.

estebanlm:
and now you need to give me a new PR fixing that, as @Ducasse merged :P

src/NewTools-SettingsBrowser/Color.extension.st
hernanmd: Thanks. I'll address the suggested changes in a new issue

src/NewTools-SettingsBrowser/SettingDeclaration.extension.st
hernanmd: Same as above, I will address the changes in a new issue.

PR #770, opened 2024-06-05
Categorise rewriter rules by pragma to allow easy extension

macta:
Simple change to allow easy way to add extra rules that can appear in the UI

### DIFF
--- a/src/NewTools-RewriterTools-Backend/StRewriterDemoRules.class.st
+++ b/src/NewTools-RewriterTools-Backend/StRewriterDemoRules.class.st
@@ -21,24 +21,14 @@ StRewriterDemoRules class >> defaultRule [
 
 { #category : 'api' }
 StRewriterDemoRules class >> demoRules [
-
-	^ {
-		  self simplifyToIfNilIfNotNilRule.
-		  self replaceAssertToAssertEmptyRule.
-		  self simpliflyConditionalToIsEmptyRule.
-		  self removeEqualsInAssertRule.
-		  self replaceDenyDenyEmptyRule.
-		  self replaceCharacterTabAsStringToStringTabRule.
-		  self simpliflyToIfEmptyIfNotEmptyRule.
-		  self removeIfNilRule.
-		  self replaceSelectToRejectRule.
-		  self removeAssignmentWithoutEffectRule.
-		  self replaceCollectFlattenToFlatCollectRule }
+ 
+	^ (Pragma allNamed: #rewriterRule) collect: [ :prg | prg method valueWithReceiver: nil]
 ]
 
 { #category : 'rules' }
 StRewriterDemoRules class >> removeAssignmentWithoutEffectRule [
-
+	<rewriterRule>
+	
 	| lhs rhs ruleName |
 	lhs := '`var := `var'.
 	rhs := ''.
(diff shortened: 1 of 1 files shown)


macta:
The ticket associate with this PR is https://github.com/pharo-project/pharo/issues/16753

macta:
How would we feel about renaming StRewriterDemoRules to StRewriterTemplateRules, as they are more than demo rules, there are some great ones in there and potentially other libraries could add to them ? Could do it in this PR or create a new issue - whatever is easier

src/NewTools-RewriterTools-Backend/StRewriterDemoRules.class.st
```diff
@@ -21,24 +21,14 @@ StRewriterDemoRules class >> defaultRule [
...
-		  self removeEqualsInAssertRule.
-		  self replaceDenyDenyEmptyRule.
-		  self replaceCharacterTabAsStringToStringTabRule.
-		  self simpliflyToIfEmptyIfNotEmptyRule.
-		  self removeIfNilRule.
-		  self replaceSelectToRejectRule.
-		  self removeAssignmentWithoutEffectRule.
-		  self replaceCollectFlattenToFlatCollectRule }
+ 
+	^ (Pragma allNamed: #rewriterRule) collect: [ :prg | prg method valueWithReceiver: nil]
```
astares, line 25: Would not be better to call the pragma #rewriterDemoRule?

macta: I was thinking it would be better to remove the Demo bit altogether as they look like legitimate rules to me - and the UI that invokes them - the ReWriteEditor calls them an example rule. So possibly example rule? Although would template make more sense? Then per my previous comment - rename the class to StRewriterTemplateRules. These are templates you can use to make code changes. I wanted to make this change so that libraries like CodeParadise could add to the list to help users update their code

Ducasse:
Yes tell me when you need that I have a look.

macta:
@Ducasse 
> Yes tell me when you need that I have a look.

Not sure what you are referring to here ? And I'm not quite sure how to complete the idea in this thread and move things forward.

I think we all agree its a good idea to make the rules more extensible - so making them pragma based is an easy enhancement to what we have.

The question is what to call the pragma (the feeling seems to be that its name should match the class holding the rules) - so should I just rename that Demo class to StRewriterTemplateRules, and the pragma <rewriterTemplateRule> all in this PR?

Happy to do that and then we can move on.

In a seperate PR I can also update the docs for Rewriter - and potentially move them from a window into the normal PharoDocs format and add a paragraph with an example from my email that really helped me better understand and pull the ideas together.

I've realised the rewriter is a much underused feature - and its very cool.

Ducasse:
Call it ruleDemo?

macta:
> Call it ruleDemo?

I think the word Demo trivialises what they do/are - many of the rules in that list are real things, you can use them now - I would say they are templates and we want projects to included templates that others should use. Thats why I think that holding class is badly named as well - particularly as the real RewriteTool links in to show them?

macta:
But if you prefer <ruleDemo> I can adjust it, and at least its done.

astares:
This topic confirms to me again: if one takes it serious the hardest part in programming is to find proper (intention revealing) names :)

Ducasse:
Then call them exampleRule

jordanmontt (approved):
Looks good to me

jordanmontt:
> The StRewriterDemoRules class (which is badly named - and should be called StRewriterSampleRules or StRewriterTemplateRules)

I agree, we can find a better name. `StRewriterSampleRules` seems better for me.

PR #772, opened 2024-06-06
Make setting pages less dense without comments for each setting

hernanmd:
This PR includes two main changes to have better setting documentation in the New Settings Browser:

- Remove explicit documentation presenter below each setting. It makes loading settings a lot faster.
- Better class names for setting presenters.

Screenshot 2024-06-06 at 16 23 39


demarey:
I like it. It is better than the previous layout but now, it misses the documentation.
As discussed, you could define the right panel with `newPanedLayoutTopToBottom:` and add the documentation at the bottom side + set the position of slider to 80 percent by example.

hernanmd:
> I like it. It is better than the previous layout but now, it misses the documentation. As discussed, you could define the right panel with `newPanedLayoutTopToBottom:` and add the documentation at the bottom side + set the position of slider to 80 percent by example.

Yes, step by step :)
I will add it in another issue/PR so I don't make this one too big.

Ducasse:
```
StSettingsBrowserTest
 ✗ #testSimulateCategorySelection (42ms)
```

PR #781, opened 2024-06-14
Deprecate asFileName and re-write senders

hernanmd:
Port this fix to Pharo 13 (re-work from #694).
Update baseline: Two new packages will contain miscellaneous utilities. 
Move #fuelOut outside the NewTools-Utils (package NewTools-Utils belongs to the Debugger group).


Ducasse:
Oops I merged but the build looks strange so I will revert.

PR #786, opened 2024-06-19
adding simple test for `StDebuggerContextInteractionModel>>#needRequestorScope:` to avoid future non-regression

adri09070:
@StevenCostiou


src/NewTools-Debugger-Tests/StDebuggerContextInteractionModelTest.class.st
```diff
@@ -216,6 +216,12 @@ StDebuggerContextInteractionModelTest >> testInteractionModelArePreservedAfterCh
 	self assert: (model hasBindingOf: 'tempVariableForTest') not
 ]
 
+{ #category : 'tests' }
+StDebuggerContextInteractionModelTest >> testNeedRequestorScope [
```
demarey, line 220: the test name does not explain the purpose of the test.
Why `needRequestorScope`should be true after initialization of the model?

PR #787, opened 2024-06-19
Fix Open Playground menu item hanging the image

hernanmd:
Fix reported and provided by @astares in https://github.com/pharo-project/pharo/issues/16778


jecisc:
It might be nice to add a smoke test

MarcusDenker (approved):

PR #790, opened 2024-06-26
Enable multiple selection in the open files/directories dialog

### LINKED ISSUE #789: Add multiple selection to open file/directory presenters
In the new file browser, we should be able to select multiple files or directories to open.

hernanmd:
Fixes #789

### DIFF
--- a/src/NewTools-FileBrowser/StFileDialogPresenter.class.st
+++ b/src/NewTools-FileBrowser/StFileDialogPresenter.class.st
@@ -320,7 +320,9 @@ StFileDialogPresenter >> open [
 StFileDialogPresenter >> openModal [
 
 	super openModal cancelled ifTrue: [ ^ nil ].
-	^ self selectedEntry
+	self isMultipleSelection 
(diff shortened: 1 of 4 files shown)


src/NewTools-FileBrowser/StFileDialogPresenter.class.st
```diff
@@ -320,7 +320,9 @@ StFileDialogPresenter >> open [
 StFileDialogPresenter >> openModal [
 
 	super openModal cancelled ifTrue: [ ^ nil ].
-	^ self selectedEntry
+	self isMultipleSelection
```
Ducasse, line 323: I was wondering if we should not use the selection.
Then I do not understand because we should not have a method returning a collection or an element. 
My impression is that we should use selectedEntries because it should return either a list with a single or multiple elements.

hernanmd: I didn't want to change it now because it could break other tools, but I will check because not many tools are using this now.

jecisc (approved):

PR #795, opened 2024-07-10
Remember debugger size and position

hernanmd:
Add lastKnownExtent and lastKnownPosition to preserve the last extent and position for the next debug request window.
Add a setting to enable this behavior (enabled by default after this PR).


StevenCostiou:
Thank you Hernan.

However I would not like to do it like that, we spent a lot of time removing similar code from the debugger.
It solves the problem (I think) but it only add unnecessary complexity to the debugger, using globals etc.

I would like to find time to discuss the solution, for instance to move all settings of the debugger into a single "debugger setting" object instead of class variables.

Please also come to the debugging team to discuss before implementing without us knowing, because we can share a bit of vision and technical constraints and objectives.

StevenCostiou:
@adri09070 can you have a look?

hernanmd:
> Thank you Hernan.
> 
> However I would not like to do it like that, we spent a lot of time removing similar code from the debugger. It solves the problem (I think) but it only add unnecessary complexity to the debugger, using globals etc.
> 
> I would like to find time to discuss the solution, for instance to move all settings of the debugger into a single "debugger setting" object instead of class variables.
> 
> Please also come to the debugging team to discuss before implementing without us knowing, because we can share a bit of vision and technical constraints and objectives.

Hi Steven,

I've updated the PR with a cleaner StDebugger "remember position" implementation by moving the settings methods to a new class, StDebuggerSettings. Please let me know of any appropriate enhancements.

hernanmd:
I've fixed some additional issues, so the CI should pass and can be reviewed "safely"

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -498,6 +498,12 @@ StDebugger >> debuggerInspectorModelClass [
 	^ StDebuggerInspectorModel
 ]
 
+{ #category : 'initialization' }
+StDebugger >> debuggerSettings [
```
StevenCostiou: the method name should be simply "settings"

StevenCostiou: it should also be class side

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -755,6 +761,12 @@ StDebugger >> initializePresenters [
 	self layout: self defaultLayout
 ]
 
+{ #category : 'initialization' }
+StDebugger >> initializeSettings: aWindowPresenter [
+
+	self debuggerSettings initializeSettings: aWindowPresenter
+]
```
StevenCostiou: the call to the settings should be inlined in the method needing it.
here there is no reason that the parameter would be a window presenter

StevenCostiou: hence I would remove that method entirely

src/NewTools-Debugger/StDebugger.class.st
```diff
@@ -814,9 +826,11 @@ StDebugger >> initializeWindow: aWindowPresenter [
 	super initializeWindow: aWindowPresenter.
 	aWindowPresenter
 		title: self debuggerActionModel statusStringForContext;
-		initialExtent: self initialExtent;
+		initialExtent: self debuggerSettings lastKnownExtent;
+		initialPosition: self debuggerSettings lastKnownPosition;
 		whenClosedDo: [ self clear ].
-
+		
+	self initializeSettings: aWindowPresenter.
```
StevenCostiou: I would inline the window extent setting here (by calling the settings)

src/NewTools-Debugger/StDebuggerSettings.class.st
```diff
@@ -0,0 +1,128 @@
...
+{ #category : 'initialization' }
+StDebuggerSettings class >> initializeSettings: aWindowPresenter [
+
+	self preserveWindowPosition ifFalse: [ ^ self ].
+	aWindowPresenter window ifNil: [ ^ self ].
+	aWindowPresenter whenClosedDo: [
+		self
+			lastKnownPosition: aWindowPresenter window position;
+			lastKnownExtent: aWindowPresenter window extent ]
+]
```
StevenCostiou: this method should not be called "initializeSettings:" but rather "configure debugger window" or something similar.

StevenCostiou (changes requested):
In addition to my comments, we need tests ensuring the new introduced behavior.

src/NewTools-Debugger/StDebugger.class.st
hernanmd: You mean?

```smalltalk
initializeWindow: aWindowPresenter

 "blabla"
self class settings initializeSettings: aWindowPresenter.

```

src/NewTools-Debugger/StDebugger.class.st
hernanmd: Done

src/NewTools-Debugger/StDebuggerSettings.class.st
hernanmd: For compatibility with Spec 2, I renamed it as `initializeWindow:`

src/NewTools-Debugger/StDebugger.class.st
hernanmd: Done

src/NewTools-Debugger/StDebugger.class.st
StevenCostiou: Rather something like `self settings setWindowPosition: aWindowPresenter` and `self settings` is implemented as `^self class settings`.

The parameter to a method named `initializeSettings:` should not be a presenter but a setting or something else.

hernanmd: Done

src/NewTools-Debugger/StDebuggerSettings.class.st
StevenCostiou: I don't know if it is an initialization really, as the usage is in the end to capture in the settings the last known position of the window.

I also think that there should not be any compatbility with spec, perhaps settings should have an explicit "setting" api.

I didn't catch it at start, but any chance of getting rid of the nil check?
Is it possible that a closing window presenter has no window?
I know sometimes spec does strange things and we can't do anything but I like to try.

StevenCostiou: thanks for the hard work :)

hernanmd:
> In addition to my comments, we need tests ensuring the new introduced behavior.

Done in the last PR commit.

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,56 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	"Now we open the debugger again"
+	self setUp.
+	dbg := (self debuggerOn: session) open.
+	"Check the debugger window is set to the new position"
+	self assert: dbg window window position equals: newDbgPosition.
+	"Set back the last position regardless of this test"
+	dbg window window position: lastUserPosition.
+
+]
+
```
adri09070: 1) I feel like each of these tests should be split into two smaller tests:
- one test that checks that the value `lastKnownExtent` (resp. `lastKnownPosition`) from `StDebuggerSettings` is changed when a debugger window is closed
- one test that when opening a new debugger, the extent (resp. position) of its window is the one registered in `StDebuggerSettings`

2) Also, it is really weird to call `#setUp` and `#tearDown` in a test. 
I think it breaks a lot of things that forces you to add weird `ifNotNil:` checks.
If the debugger needs to be cleared, then you could extract  from the `#tearDown` a method that clears the debugger and that you could call in your tests, instead of calling `tearDown`  and `setUp`

3) When setting back the values of  `lastKnownExtent` (resp. `lastKnownPosition`), I think that you should use the setters of `StDebugger` settings in an `ensure:` block

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,56 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	"Now we open the debugger again"
+	self setUp.
+	dbg := (self debuggerOn: session) open.
+	"Check the debugger window is set to the new extent"
+	self assert: dbg window window extent equals: newDbgExtent.
+	"Set back the last extent regardless of this test"
+	dbg window window extent: lastUserExtent.
+
+]
```
adri09070: this test is failing

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,56 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg close.
+	self tearDown.
+
+	"Now we open the debugger again"
+	self setUp.
+	dbg := (self debuggerOn: session) open.
+	"Check the debugger window is set to the new position"
+	self assert: dbg window window position equals: newDbgPosition.
+	"Set back the last position regardless of this test"
+	dbg window window position: lastUserPosition.
```
adri09070: maybe we should add tests that test the behavior when `StDebuggerSettings class>>#PreserveWindowPositionAndExtent` is false

adri09070 (changes requested):

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
hernanmd: Which Pharo version do you use?

Using the latest Pharo 13.0.0 here, and it works:
Build information: Pharo-13.0.0+SNAPSHOT.build.124.sha.d1dbcc0440c7d27e9a05e6ea2a659f5ddb41a484 (64 Bit)

Screenshot 2024-07-15 at 15 53 55

adri09070: the last P13 build (look at the CI)

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
hernanmd: I agree with adding two tests. But please note that they are not the same. Opening the debugger window and then programmatically changing its position ensures that we test real user interaction with the window positioning. So, the position change should be updated **from the interaction**, not just from the global, and it should match the position set in the test. 

If you do not open an actual debugger window, you check if a global variable was updated accordingly.

Regarding the manual calling of `tearDown` and `setUp`, did you try to execute the test without sending `tearDown` and `setUp`? When I tried, many things started to fail in the debugger's cleanup.

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
hernanmd: I see, I don't understand why is it failing in the CI.

I would rewrite the test following your suggestions, although I think it doesn't reflect the actual user interaction.

hernanmd:
Guys, yesterday this PR was green and today I've seen some conflicts so it would be nice to have it integrated on time before other PR's make it obsolete, otherwise, I have to work on this forever.

Now I'm checking this again because StDebugger initializeWindow: was changed somewhere.

hernanmd:
The failing test is unrelated.

hernanmd:
> Guys, yesterday this PR was green and today I've seen some conflicts so it would be nice to have it integrated on time before other PR's make it obsolete, otherwise, I have to work on this forever.
> 
> Now I'm checking this again because StDebugger initializeWindow: was changed somewhere.

Done. Could you have a look @StevenCostiou ?

StevenCostiou:
Perhaps tomorrow, I am not available today.
But worry not, the debugger code does not change that often, that is just bad luck.

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
 	self assert: debuggerContext context identicalTo: ctx
 ]
 
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testOpenDebuggerWithLastKnownExtent [
+
+	| lastUserPreservation |
+
+	lastUserPreservation := self settings preserveWindowPositionAndExtent.
```
adri09070: this should be done in `setUp`. If the test fails, then this settings won't be set to how it used to be

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg := (self debuggerOn: session) open.
+	"Check window is opened with the saved extent"
+	self 
+		assert: self settings lastKnownExtent 
+		equals: dbg window window extent.
+
+	dbg close.
+
+	"Restore the original preservation setting"		
+	self settings preserveWindowPositionAndExtent: lastUserPreservation.
```
adri09070: this should be done in `tearDown`, for the same reasons

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testOpenDebuggerWithLastKnownPosition [
+
+	| lastUserPreservation |
+
+	lastUserPreservation := self settings preserveWindowPositionAndExtent.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg := (self debuggerOn: session) open.
+	"Check window is opened with the saved position"
+	self 
+		assert: self settings lastKnownPosition
+		equals: dbg window window position.
+
+	dbg close.
+
+	"Restore the original preservation setting"		
+	self settings preserveWindowPositionAndExtent: lastUserPreservation.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testPreserveLastKnownExtentIsFalse [
+
+	| newDbgExtent lastUserPreservation |
+
+	lastUserPreservation := self settings preserveWindowPositionAndExtent.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg window window extent: newDbgExtent.
+	"New extent is saved when the window is closed"
+	dbg close.
+
+	self 
+		deny: self settings lastKnownExtent 
+		equals: newDbgExtent.
+
+	"Restore the original preservation setting"		
+	self settings preserveWindowPositionAndExtent: lastUserPreservation.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testPreserveLastKnownExtentIsTrue [
+
+	| newDbgExtent lastUserExtent |
+
+	dbg := (self debuggerOn: session) open.
+	"First we save the user extent (test running should not modify user changes)"
+	lastUserExtent := self settings lastKnownExtent.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg window window extent: newDbgExtent.
+	"New extent is saved when the window is closed"
+	dbg close.
+
+	self 
+		assert: self settings lastKnownExtent 
+		equals: newDbgExtent.
+
+	"Restore the original extent"		
+	self settings lastKnownExtent: lastUserExtent.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testPreserveLastKnownPositionIsFalse [
+
+	| newDbgPosition lastUserPreservation |
+
+	lastUserPreservation := self settings preserveWindowPositionAndExtent.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg window window position: newDbgPosition.
+	"New extent is saved when the window is closed"
+	dbg close.
+
+	self 
+		deny: self settings lastKnownExtent 
+		equals: newDbgPosition.
+
+	"Restore the original preservation setting"		
+	self settings preserveWindowPositionAndExtent: lastUserPreservation.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testPreserveLastKnownPositionIsTrue [
+
+	| newDbgPosition lastUserPosition |
+
+	dbg := (self debuggerOn: session) open.
+	"First we save the user position (test running should not modify user changes)"
+	lastUserPosition := self settings lastKnownPosition.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -752,6 +759,136 @@ StDebuggerTest >> testNewDebuggerContextFor [
...
+	dbg window window position: newDbgPosition.
+	"New position is saved when the window is closed"
+	dbg close.
+
+	self
+		assert: newDbgPosition 
+		equals: self settings lastKnownPosition.
+		
+	"Restore the original position"		
+	self settings lastKnownPosition: lastUserPosition.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -1216,3 +1353,47 @@ StDebuggerTest >> testUpdateSourceCodeFor [
...
 ]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testValueOfLastKnownExtentIsChangedWhenWindowIsClosed [
+
+	| newDbgExtent lastUserExtent|
+
+	dbg := (self debuggerOn: session) open.
+	"First we save the user position (test running should not modify user changes)"
+	lastUserExtent := self settings lastKnownExtent.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -1216,3 +1353,47 @@ StDebuggerTest >> testUpdateSourceCodeFor [
...
+	dbg window window extent: newDbgExtent.
+	"New position is saved when the window is closed"
+	dbg close.
+
+	self 
+		assert: self settings lastKnownExtent
+		equals: newDbgExtent.
+
+	"Restore the original extent"		
+	self settings lastKnownExtent: lastUserExtent.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -1216,3 +1353,47 @@ StDebuggerTest >> testUpdateSourceCodeFor [
...
+]
+
+{ #category : 'tests - context inspector' }
+StDebuggerTest >> testValueOfLastKnownPositionIsChangedWhenWindowIsClosed [
+
+	| newDbgPosition lastUserPosition |
+
+	dbg := (self debuggerOn: session) open.
+	"First we save the user position (test running should not modify user changes)"
+	lastUserPosition := self settings lastKnownPosition.
```
adri09070: should be done in `setUp`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -1216,3 +1353,47 @@ StDebuggerTest >> testUpdateSourceCodeFor [
...
+	dbg window window position: newDbgPosition.
+	"New position is saved when the window is closed"
+	dbg close.
+
+	self 
+		assert: self settings lastKnownPosition 
+		equals: newDbgPosition.
+		
+	"Restore the original position"		
+	self settings lastKnownPosition: lastUserPosition.
```
adri09070: should be done in `tearDown`

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -121,10 +127,11 @@ StDebuggerTest >> tearDown [
...
 	StTestDebuggerProvider removeSelector: #foobar.
-	oldMethodToSourceDictionary associationsDo: [ :assoc |
- 		assoc key methodClass
- 			compile: assoc value
- 			classified: assoc key protocolName ].
+	oldMethodToSourceDictionary ifNotNil: [ 
+		oldMethodToSourceDictionary	associationsDo: [ :assoc |
+ 			assoc key methodClass
+ 				compile: assoc value
+ 				classified: assoc key protocolName ] ].
```
adri09070: this `ifNil:` call  should really be unnecessary

src/NewTools-Debugger-Tests/StDebuggerTest.class.st
```diff
@@ -103,7 +109,7 @@ StDebuggerTest >> subscriptionsForStDebuggerExtensionActivationToggle [
 { #category : 'running' }
 StDebuggerTest >> tearDown [
 
-	session clear.
+	session ifNotNil: [ session clear ].
```
adri09070: this `ifNil:` call should really be unnecessary

adri09070 (changes requested):
Your test fails randomly and I can't exactly explain why but here are things that should be changed in your tests and that will make your test pass everytime

StevenCostiou:
For me now there are two problems left:
- what about those nil checks? can we do something?
- tests are all red (I didn't check what's happening)
If those are fixed, can we agree to merge?

adri09070:
Concerning the failing tests, they are not related

adri09070 (approved):

hernanmd:
> For me now there are two problems left:
> 
>     * what about those nil checks? can we do something?
> 
>     * tests are all red (I didn't check what's happening)
>       If those are fixed, can we agree to merge?

I removed the ifNil: checks since I need to work on other issues.

PR #796, opened 2024-07-12
add error handler for latin1 encoding failure

tomooda:
When inspecting a unicode string that can't be encoded in Latin-1, the encoding tab of the inspector fails because the encoder emits error. This PR adds an error handler so that encoding error will result in nil.


Ducasse:
Thanks!

PR #797, opened 2024-07-16
Fixes critics browser problems

jecisc:
This change fixes 3 main problems.

1) SpMorphicDiffUnifiedAdapter was using a morph from this package. I'm moving it to Spec in this PR: https://github.com/pharo-spec/Spec/pull/1570 So I'm removing it from here
2) StRewriteCritiqueChangesBrowserPresenter was using a morph directly, now it uses a presenter instead
3) StRewriteCritiqueChangesBrowserPresenter was inheriting from a deperecated class. For now I changed the superclass and I copy pasted the methods needed from the superclass but maybe we can do simpler in the future. But this is a first step to be able to remove the deprecated super classes.


Ducasse:
Thanks a lot! This is great to see this making progress.

MarcusDenker (approved):

MarcusDenker:
Just one not related failure:

```
ReleaseTest
 ✗ #testUndeclared (115ms)
TestFailure: Found undeclared Variables: 
StClassCritiqueRuleSelectorPresenter in: StClassCritiqueBrowserPresenter class>>#openOnCurrentWorkingConfiguration
StSystemReporter in: SmalltalkCIPharo9>>#imageInfo
```

PR #800, opened 2024-07-18
Remove now unused CBUnifiedDiffJoinSection

jecisc:
This class is now in Pharo under another name


demarey:
`CBUnifiedDiffJoinSection`has been renamed `UnifiedDiffChangesMorph` and is packaged in `Tool-Diff`

PR #805, opened 2024-07-18
Create NewTools-Calypso

jecisc:
This package is empty for now and here is the reason:

I'm introducing this package because I migrated a part of Calypso in NewTools-Core. But this introduces a dependency on some packages we do not want in the core of Pharo. So, to fix this dependency here is my proposal:
- Introduce an empty package NewTools-Calypso with a group 'Calypso'
- Load this new group in Pharo before Calypso but after NewTools-Core
- Move the presenter used by Calypso to this new package and cut the bad dependency


jecisc:
I think there is simpler to do in the end. I'll try something else

PR #806, opened 2024-07-18
Fix dependency test

jecisc:
A dependency test of Pharo is failing because of a Commander 1 dependency.

I added the possibility to handle Commander2 errors in Calypso and here I am migrating the errors from Commander1 to Commander2 to fix the dependency failure


MarcusDenker:
failing tests not related

MarcusDenker (approved):

PR #807, opened 2024-07-19
Adding command to add breakpoints in the debugger

adri09070:
Needs https://github.com/pharo-project/pharo/pull/16919 to be merged first

### DIFF
--- a/src/NewTools-Debugger-Commands/StBreakOnStackCommand.class.st
+++ b/src/NewTools-Debugger-Commands/StBreakOnStackCommand.class.st
@@ -1,0 +1,40 @@
+"
+I am a debugger command class that allows to select an instruction in the debugger to dynamically install a breakpoint on the corresponding AST node.
+"
+Class {
+	#name : 'StBreakOnStackCommand',
+	#superclass : 'StDebuggerCommand',
+	#category : 'NewTools-Debugger-Commands',
+	#package : 'NewTools-Debugger-Commands'
+}
+
+{ #category : 'default' }
+StBreakOnStackCommand class >> defaultDescription [
+
+	^ 'Recompile method to add a breakpoint before selection and replace the compiled methods on stack by its reflective version'
+]
+
+{ #category : 'initialization' }
+StBreakOnStackCommand class >> defaultIconName [
+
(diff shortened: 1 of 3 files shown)


adri09070:
@StevenCostiou

StevenCostiou:
thanks I will look at this

src/NewTools-Debugger-Commands/StBreakOnStackCommand.class.st
```diff
@@ -0,0 +1,40 @@
...
+StBreakOnStackCommand class >> defaultIconName [
+
+	^ #breakOnStack
+]
+
+{ #category : 'default' }
+StBreakOnStackCommand class >> defaultName [
+
+	<toolbarDebugCommand: 50>
+	^ 'Break on stack'
```
StevenCostiou: I would name it just "break"

src/NewTools-Debugger-Commands/StBreakOnStackCommand.class.st
```diff
@@ -0,0 +1,40 @@
...
+{ #category : 'default' }
+StBreakOnStackCommand class >> defaultDescription [
+
+	^ 'Recompile method to add a breakpoint before selection and replace the compiled methods on stack by its reflective version'
+]
+
+{ #category : 'initialization' }
+StBreakOnStackCommand class >> defaultIconName [
+
+	^ #breakOnStack
```
StevenCostiou: Does this icon exists?
Why not #smallDebug for example?

StevenCostiou (changes requested):

adri09070:
done

StevenCostiou:
The contributions are good.
We can go when tests are fixed.
There is only one related failing test, apparently there is a implemented selector that is never sent?

adri09070:
I think this is the message send `#findContextsForMethod:`, which is implemented in the other PR that should be merged first. I saw that there is an `assert:` left that I will remove

adri09070:
We should wait a few hours so that the build including my changes of https://github.com/pharo-project/pharo/pull/16919 has been built

adri09070:
For me, it is good to merge

PR #821, opened 2024-08-09
Add BrowserSearch dialog for Calypso

jecisc:
This introduce a new search presenter to use instead of the Morphic search browser of Calypso.

If this gets integrated I have a PR for Pharo to remove all users of the Morphic browser search dialog


Ducasse:
The broken test is unrelated.

Issue #808, opened 2024-07-19
StFileOrFolder open should take the input as a possible selection


Ducasse:
When I past a path of a given folder or file for example using the green pencil on the top right
for example /Users/ducasse/Workspace/FirstCircle/Writing/Working/researchnotes in the screenshot
researchnotes should be proposed as a selection. 

Else I have to navigate on the parent, then scroll and look in the children of the parent, just to select the correct child here researchnotes

Ducasse:
<img width="1066" alt="Capture 2024-07-19 at 21 28 20" src="https://github.com/user-attachments/assets/92864f44-077f-4fd2-a0e8-ad1748e9c739">

Ducasse:
I'm in Pharo 12 so may be I should move to Pharo 13

hernanmd:
I'm looking into this now, but I'm not sure if I understand exactly what you mean. Could you please paste a screenshot of the result view you expect?

Because we have multiple selections here. One is tree navigation selection:

<img width="1439" alt="Screenshot 2024-08-17 at 15 16 41" src="https://github.com/user-attachments/assets/c2bf9549-9c98-43f2-a114-d7f20d600415">

And the edit path presenter **enters** the pasted directory. Do you mean the selection presenter (in the red box below)?

<img width="1432" alt="Screenshot 2024-08-17 at 15 21 28" src="https://github.com/user-attachments/assets/78fd60cf-f5f4-474b-a7d2-049a66f84710">

Ducasse:
Hi hernan I can show you what I mean monday or tuesday.

Ducasse:
I meant the selection that in red on your screen. 
Else I have to navigate on the parent, then scroll and look in the children of the parent, just to select the correct child here researchnotes

hernanmd:
> I meant the selection that in red on your screen. Else I have to navigate on the parent, then scroll and look in the children of the parent, just to select the correct child here researchnotes

Ok, I pushed a PR in #822 :)

PR #822, opened 2024-08-17
Fix missing updates in File Browser file/directory name presenter

hernanmd:
Fix StFileOrFolder open should take the input as a possible selection, as reported by @Ducasse in #808.


hernanmd:
Failing because of the SmalltalkCI Author issue: "KeyNotFound: key #Author not found in SystemEnvironment"

demarey (approved):

PR #832, opened 2024-09-04
Inprove the UX of the Calypso search

jecisc:
When we navigate up and down in the search dialog, we should scroll to the selected element.

I also commited some deprecation rewrites


hernanmd (commented):
Changes are ok for me

PR #834, opened 2024-09-08
Add setting browser tabs

hernanmd:
This PR includes a major upgrade in the new settings browser UI to include tabs for 2nd-level categories.
Each 2nd-level category now includes a "General" settings tab, which holds those settings that do not belong to a sub-category.
Additionally, each sub-category (or tab) has a "General" sub-section to hold its general settings.

https://github.com/user-attachments/assets/46b07cb4-f834-47b8-9d8a-b8a499f73747


Ducasse:
Good. We should finish this setting browser :)

PR #841, opened 2024-09-19
Message Browser advances (scoping)

hernanmd:
This PR includes features to complete a minimum usable implementation of the new Message Browser, now implemented mostly in `StMessageListPresenter` and `StMethodToolbarPresenter`. 

Most of the changes are related to adding the features of the "scopes drop list" which is initially opened with the scope of the queried method, and as the user selects methods is dynamically populated with selected packages, classes, and hierarchies. This way the user can select new scopes to filter the list of messages.
 
## Demos

- Dynamically populating packages and switching to package scopes:

https://github.com/user-attachments/assets/95234632-49ef-4e38-947c-a5fac79c54f9

- Selection of multiple methods to query its senders. Scrolling speed. Opening new Implementors window from selection.

https://github.com/user-attachments/assets/ec738b4d-765e-4bdc-8b63-943927ac82f2

- Switching scope to a custom scope

https://github.com/user-attachments/assets/b1dbf4e1-4c03-4a36-b749-99101018a5df

## Notes

- For packages: The system browser is not needed anymore to create package scopes (`ClyPackageScope`), i.e.: `ClyQueryBrowserMorph>>#packageScopeOfSelectedItems`
- For classes: Getting class information no longer needs `systemDefinition` from `ClyDataSourceItem>>#systemDefinition` and `ClyBrowserItem>>#systemDefinition`.
- For class hierarchies: Not inherited scopes and annotations used to create them, `ClyInheritedScopeProviderAnnotation` is unnecessary. 
- Many of the methods of `ClyQueryBrowserMorph` were rewritten in Spec 2 NewTools, to detach the scoping functionality from the Calypso query system, which is also attached to Morphic. This also reduced the complexity of the implementation.
- The new scopes drop list includes the result of `ScopesManager availableScopes` to filter composite environments defined in the Scopes Editor.
- The title of the message list browser includes **all** the methods selected for both sender and implementor windows.

## Known issues

These issues could be fixed in upcoming PR's

- The hierarchy scopes are duplicated in the scopes list.
- New instances of the message browsers will be opened in new windows, instead of spawning as tabs.

### DIFF
--- a/src/NewTools-MethodBrowsers/RBBrowserEnvironment.extension.st
+++ b/src/NewTools-MethodBrowsers/RBBrowserEnvironment.extension.st
@@ -1,0 +1,8 @@
+Extension { #name : 'RBBrowserEnvironment' }
+
+{ #category : '*NewTools-MethodBrowsers' }
+RBBrowserEnvironment >> selectMessagesFrom: aCollection [ 
+	"Since the receiver is a system environment, answer all messages in aCollection"	
+
+	^ aCollection 
+]
--- a/src/NewTools-MethodBrowsers/RBBrowserEnvironment.extension.st
+++ b/src/NewTools-MethodBrowsers/RBBrowserEnvironment.extension.st
@@ -1,8 +1,8 @@
 Extension { #name : 'RBBrowserEnvironment' }
 
 { #category : '*NewTools-MethodBrowsers' }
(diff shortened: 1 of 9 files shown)


Ducasse:
Super cool!
Tx a lot Hernan!

PR #846, opened 2024-09-21
SycCmClassCritiqueCommand should be in Extras submenu

hernanmd:
One-liner PR to fix "Class Critique Browser menu Item should not be in refactoring" reported in https://github.com/pharo-project/pharo/issues/17079


hernanmd:
Thanks!

PR #852, opened 2024-09-27
Remove senders of scaledByDisplayScaleFactor and displayScaleFactor

guillep:
Remove senders of scaledByDisplayScaleFactor and displayScaleFactor.

This was superseeded by new scale factor support superseeded by new scale factor support


Ducasse:
Refactoring tests are broken but not due to this PR

PR #854, opened 2024-09-27
Correct evaluation of dynamic variables from the debugger

### LINKED ISSUE #17038: Debugger doesn’t apply dynamic variable values from debugged context to evaluations
Using Pharo 13 build 193, on the following screenshot, I put a halt in a snippet that sets a value for [TestDynamicVariable](https://github.com/pharo-project/pharo/blob/e988133ca43cbd49cbae9314e0a3683f5ee794a2/src/Kernel-Tests/TestDynamicVariable.class.st), then in the debugger I try to inspect the send of `#value` to that variable:

<p align="center">

</p>

The result is nil. It would be nice if the debugger applied the dynamic variable values from the debugged context to the evaluation, so that the result would be 42.

For a more concrete example, [load Seaside](https://github.com/SeasideSt/Seaside#load-seaside) and put a breakpoint at the send of `#rootClass` in [`#createRoot` on WAInitialRenderLoopContinuation](https://github.com/SeasideSt/Seaside/blob/0f7d6f710c187be530b5e94d7b878138bbb28852/repository/Seaside-RenderLoop.package/WAInitialRenderLoopContinuation.class/instance/createRoot.st), then open a web browser on http://localhost:8080/ and try inspecting the send of `#rootClass` in the debugger that is opened, you’ll get a second debugger rather than an inspector as a WARequestContextNotFound is signaled due to the dynamic variable WACurrentRequestContext’s value being nil.

StevenCostiou:
Fixes https://github.com/pharo-project/pharo/issues/17038


StevenCostiou:
hmm something is wrong in the tests

StevenCostiou:
no idea why I have an error on newtools but not on pharo-integration...

StevenCostiou:
So the code works in the debugger but the test is flaky.
This is due to the means to build the test, it is not easy.

StevenCostiou:
Trying to improve tests, but I think it is stiill flaky.
I recommend merging the main commit but not the test, as at worst cases the behavior would be the same as before.

StevenCostiou:
Ok it seems to pass now...

MarcusDenker (approved):

StevenCostiou:
Ok the failing tests are because of the change.
I will have to look for side effects.

StevenCostiou:
I think the remaining failing test is not due to my changes (I can be wrong).
I have not added code that would be identical to methods in super classes.

StevenCostiou:
Ok this seems to be reproducible on the CI.
Will check again, but in my image this release test does pass.

StevenCostiou:
The failing tests are in new tools now, but concern the refactorings...
I cannot udnerstand the relation.

And its working on Pharo...

StevenCostiou:
We should merge @MarcusDenker

PR #861, opened 2024-10-11
858 debugger add an inspect button to the actual vs expected view

### LINKED ISSUE #858: [Debugger] Add an inspect button to the Actual VS Expected view
I have a case where I assert that the result of two methods are identical. They return strings but they have different invisible characters. 

It would have helped we if I could have gotten an "inspect" button next to the Actual and Expected view doing a text diff at the bottom of the debugger

StevenCostiou:
Fixes #858

### DIFF
--- a/src/NewTools-Debugger-Extensions/TestFailure.extension.st
+++ b/src/NewTools-Debugger-Extensions/TestFailure.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : 'TestFailure' }
+
+{ #category : '*NewTools-Debugger-Extensions' }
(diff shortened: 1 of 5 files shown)


Ducasse:
Super nice to see more care about the debugger.

StevenCostiou:
Friday is debugger day :)

StevenCostiou:
@MarcusDenker the release test is the one we spoke about, is being worked on in this PR https://github.com/pharo-spec/NewTools/pull/854
Everything is fine on the full Pharo.
Can we try a merge and see?

PR #870, opened 2024-10-21
sort playground pages by modification time

estebanlm:
pages were shown in creation time which is dumb. Let's use modification time which will sort them better. Also, refatcor layout to instance side since class side is old and not recommended anymore.


MarcusDenker:
Two tests are failing when running all tests that I have not seen failing for the main CI recently:

```
testCheckExistenceOfNonExistingFileDoesNotThrowException
testLinkOneShot
```

test for newtools are green

MarcusDenker (approved):
ok

jecisc (approved):
Nice!

PR #873, opened 2024-10-28
Add search feature Settings Browser

### LINKED ISSUE #17206: StSettingsApplication should select the first pane when we select a topics
When we click on code browsing we should not see this. 

Capture 2024-10-03 at 21 59 37

Can we move on and remove the old setting browser. This starts to take too long.

hernanmd:
This PR includes the first implementation of a live search in the New Settings Browser. 

### Details 

- It works by configuring a results tree, similar to the whole settings tree, based on the typed search input. 
- The minimum number of characters to start a search is currently 3, but could be easily made configurable.
- It also auto-selects the first tab on setting category selection. Fixes https://github.com/pharo-project/pharo/issues/17206
- Setting node search was enhanced by querying each node label by substrings, then lowercase each one and comparing it against the lowercase input pattern.


src/NewTools-SettingsBrowser/SpNotebookPresenter.extension.st
```diff
@@ -0,0 +1,19 @@
...
+SpNotebookPresenter >> hasPages [
+
+	^ self pages isNotEmpty
+]
+
+{ #category : '*NewTools-SettingsBrowser' }
+SpNotebookPresenter >> pageNamed: aTitle [
+
+	^ self pages detect: [ : page | page title = aTitle ]
+]
```
jecisc: I'd prefer this to be directly in the Notebook presenter

hernanmd: Agree. Done in https://github.com/pharo-spec/Spec/pull/1636

hernanmd:
There is a single failing test in the NewTools-Pharo integration but it seems unrelated `ReflectivityControlTest>>#testLinkOneShot` and I cannot reproduce it in the UI image:

Screenshot 2024-11-06 at 10 11 06

jecisc (approved):

PR #876, opened 2024-11-01
Fixed linux terminal command, unified method for Windows

### LINKED ISSUE #875: FileBrowser: Open in terminal ends up in error on Linux
On Linux, when a user clicks on Open in terminal from a directory context menu, it ends up in an `Error: Improper store into indexable object`. The issue is in method `StFileBrowserOpenTerminalCommand>>#openLinuxTerminalShellCommand:` where it attemps to use a `Path` as `putOn:` receiver

JanBliznicenko:
Linux open terminal command no longer ends up in an error because of improper use of Streams.
Fixes https://github.com/pharo-spec/NewTools/issues/875

Unrelated to the issue, windows variant of the command generating method is now named similarly to others.


jecisc (approved):
Thanks!

PR #877, opened 2024-11-02
Use StPharoApplication instead of custom settings application

hernanmd:
Includes the changes from the add_search_feature PR #873


jecisc:
Can you merge development into this branch to see only the changes of this PR and not the other one please?

hernanmd:
> Can you merge development into this branch to see only the changes of this PR and not the other one please?

Done

src/NewTools-SettingsBrowser-Tests/StSettingsBrowserTest.class.st
```diff
@@ -13,9 +13,8 @@ Class {
 { #category : 'running' }
 StSettingsBrowserTest >> application [
 
-	^ StSettingsApplication new
-		  useBackend: #Morphic;
-		  yourself
+	^ StPharoApplication current
```
jecisc: I am really not sure, but shouldn't we do something like `StSettingBrowser new application`? Like this, if the app changes one day, it'll still be the right one? (As I said, I'm really not sure here)

jecisc (approved):

jecisc:
✗ #testExternalIDEDependencies (73ms)
TestFailure: Given Collections do not match!
	additions : #('NewTools-FileBrowser')
	missing: #()

hernanmd:
> ✗ #testExternalIDEDependencies (73ms) TestFailure: Given Collections do not match! additions : #('NewTools-FileBrowser') missing: #()

Is this failure related to this PR? 

I don't see how it could be since I didn't add FileBrowser class references.

jecisc:
It indeed comes from the current Pharo build :(

hernanmd:
Thanks

PR #878, opened 2024-11-06
Explicit morphic/uimanager dependencies

guillep:
I propose to make all dependencies to morphic/uimanager explicit.

Probably this PR should not be "accepted", but it shows a problem: the points modified in this PR should probably pass through the application?


Ducasse:
Guille for the stFinderExampleSearch I passed the application so and now the code should be self application inform:

Ducasse:
I fixed the conflict

PR #880, opened 2024-11-06
Register new File Browser in tools registry

hernanmd:
Register the new file browser in the tools registry after Baseline load.


src/NewTools-FileBrowser/String.extension.st
```diff
@@ -0,0 +1,12 @@
+Extension { #name : 'String' }
+
+{ #category : '*NewTools-FileBrowser' }
+String >> asFileName [
```
jecisc: Is it wanted to have this? I think this kind of methods were removed from Pharo not so long ago?

hernanmd: I don't know, but I've checked and two external users use it:

- `Object>>#fuelOut`
- `CodeExporter class>>#writeSourceCodeFrom:baseName:isSt:`

Captura de pantalla 2024-11-07 a las 15 52 47

We can include it for now and open another issue to see if it is needed.

src/BaselineOfNewTools/BaselineOfNewTools.class.st
```diff
@@ -231,7 +231,8 @@ BaselineOfNewTools >> packageRepositoryURLForSpec: spec [
 { #category : 'actions' }
 BaselineOfNewTools >> postload: loader package: packageSpec [
 
-	MCMethodDefinition initializersEnabled: initializersEnabled
+	MCMethodDefinition initializersEnabled: initializersEnabled.
+	Smalltalk tools fileListTool: StFileSystemPresenter.
```
Ducasse, line 235: please use self class environment no direct reference to Smalltalk!

jecisc: The problem is that it is not the environment that holds the tool list so we cannot access it like this

jecisc:
Closing in favor of: https://github.com/pharo-spec/NewTools/pull/885

PR #882, opened 2024-11-08
Multiple FileBrowser directory handling fixes for Windows

JanBliznicenko:
Mulitiple FileBrowser fixes for Windows:
- Directory tree now scrolls to current directory (used to only work for Linux)
- Drives without any partitions are no longer listed amongst drives (a Windows 11 update started assinging drive letters to such drives)
- Breadcrumbs no longer contain Root part on Windows (used to result in error when used as there is no real single root)
- Unified drive icons between "directories" and bookmarks
- Bookmarks now (re)use already existing code for finding roots/drives instead of using own implementation

And a fix for all platforms:
- Working directory bookmark now opens the actual working directory instead of root

### DIFF
--- a/src/NewTools-FileBrowser/StDirectoryTreePresenter.class.st
+++ b/src/NewTools-FileBrowser/StDirectoryTreePresenter.class.st
@@ -49,11 +49,13 @@ StDirectoryTreePresenter >> expandPath: aFileLocator [
 	"Expand the receiver's tree to aFileLocator reference"
 
 	| path aPathForSpec currentNode |
-	path := aFileLocator asPath segments asOrderedCollection.
-	aPathForSpec := OrderedCollection new.
-	aPathForSpec add: 1.
+	path := aFileLocator absolutePath segments asOrderedCollection.
+	aPathForSpec := OrderedCollection with: 1.
 
 	currentNode := directoryTreePresenter roots anyOne.
+	Smalltalk os isWindows ifTrue: [
+		currentNode := currentNode asFileReference parent.
+		aPathForSpec := OrderedCollection new ].
 
 	path do: [ :aPart |
 		| subdirs |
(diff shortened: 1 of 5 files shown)


Ducasse:
Oh thanks a lot!!!!

src/NewTools-FileBrowser/StDirectoryTreePresenter.class.st
```diff
@@ -49,11 +49,13 @@ StDirectoryTreePresenter >> expandPath: aFileLocator [
...
 
 	| path aPathForSpec currentNode |
-	path := aFileLocator asPath segments asOrderedCollection.
-	aPathForSpec := OrderedCollection new.
-	aPathForSpec add: 1.
+	path := aFileLocator absolutePath segments asOrderedCollection.
+	aPathForSpec := OrderedCollection with: 1.
 
 	currentNode := directoryTreePresenter roots anyOne.
+	Smalltalk os isWindows ifTrue: [
```
Ducasse, line 56: Jan what do think about the following:
- could we pass the os (self class environment os) to the presenter when it is instantiated. Like that we do not have this global variable right in the middle. 
- Potentially later we could have a strategy for each os but this is another step.

hernanmd:
I'm trying to review but the commit is too behind the latest Pharo13 branch. @JanBliznicenko could you update your branch by merging the Pharo13 latest changes?

src/NewTools-FileBrowser/StDirectoryTreePresenter.class.st
JanBliznicenko: > could we pass the os (self class environment os) to the presenter when it is instantiated. Like that we do not have this global variable right in the middle.

You are right, I am, however, failing to find a way to do that without hampering usability. Problem is that most of these presenters need to know the OS before `initializePlatform`, therefore before `initialize`, and there are many methods sending it.
`new, owner:, newApplication:, newApplication:model:, newApplication:owner: newApplication:owner:model:, owner:on:`, these all are senders of `initialize`, so they could not be used and would need to have a variant with another parameter (the platform) instead.
Maybe to not actually show the contents until the platform is set and then reload the layout? IDK

JanBliznicenko:
> I'm trying to review but the commit is too behind the latest Pharo13 branch. @JanBliznicenko could you update your branch by merging the Pharo13 latest changes?

True, it seems I took quite long finishing it. I have rebased it on current commit, however, some seemingly unrelated tests seem to be failing now.

src/NewTools-FileBrowser/StDirectoryTreePresenter.class.st
hernanmd: Shouldn't the SystemEnvironment have an `os` method?
I would let this to be addressed in another issue, since it is part of a wider problemn that could need other integrations before.

hernanmd (approved):
Changes are ok. Approved

PR #883, opened 2024-11-09
Make sure that StFinder uses application for inform:

Ducasse:
- Make sure that StFinder uses application for inform:
- Pass the application to the strategy
- Now the application was a NullApplication so 
- 
- Add a subclass of SpPresenterWithModel StPresenterWithModel
- Extract a trait from StPresenter
- Apply the trait to StPresenterWithModel
- Make sure that StFinder is a subclass of StPresenterWithModel

### DIFF
--- a/src/NewTools-Core/StPresenterWithModel.class.st
+++ b/src/NewTools-Core/StPresenterWithModel.class.st
@@ -1,6 +1,6 @@
 Class {
-	#name : 'StPresenter',
-	#superclass : 'SpPresenter',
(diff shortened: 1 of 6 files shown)


estebanlm:
I like the extraction of `TStPresenter` :)
but it need to be called `StTPresenter` to follow the naming conventions.

Ducasse:
Ok I will rename it in another PR.

Ducasse:
Strange I do not see the code in latest P13 may be because it was not rebuilt.

PR #885, opened 2024-11-12
Register new File Browser in tools registry

jecisc:
Alternative to https://github.com/pharo-spec/NewTools/pull/880

I'm not putting this in the post load of the baseline because if we load a group of the baseline we do not know if the file browser will be loaded or not. Pharo seems to load a subgroup of NewTools first and the file list is not in this subgroup.


Ducasse:
The broken test is unrelated so I integrated it.

Ducasse:
Now the question is why StApplication does not have a tool registry because it should and we should register there.

PR #897, opened 2024-11-14
Restore Settings Browser missing custom styles

### LINKED ISSUE #896: Restore Setting Browser styles
## Description

A recent PR (https://github.com/pharo-spec/NewTools/pull/877/files) has changed the way we get the custom styles in the New Settings Browser.
We should ensure the Setting Browser presenters get its custom styles upon opening the UI.

hernanmd:
Fixes #896


src/NewTools-SettingsBrowser/StSettingsAbstractPresenter.class.st
```diff
@@ -23,6 +23,13 @@ StSettingsAbstractPresenter class >> appearanceIconContents [
...
 ]
 
+{ #category : 'accessing' }
+StSettingsAbstractPresenter class >> application [
```
jecisc: Shouldn't the class inherit from StPresenter to not have to do this kind of things?

hernanmd: Thanks. I've updated the PR

jecisc (approved):

PR #901, opened 2024-11-16
Configure Setting Browser initialization properties

hernanmd:
This PR includes a little fix to configure the Setting Browser main presenter initialization before other presenters initialization. This is needed because sending `initialize` before configuring the styles refers to styling elements that aren't available until later.


hernanmd:
Failing test `testLinkOneShot` is not related

Ducasse:
Tx hernan!

Ducasse:
I'm about to push a tools registration in StPharoApplication and after I will start migrating the tool to use it.

PR #912, opened 2024-11-26
Fixes #17443 on pharo

AlexisCnockaert:
Removed suggestionlist selection in initialization because it was impossible to scroll if the prtocol name changes everytime we select a protocol in the list


Ducasse:
Hi alexis 
I could not get the idea of this PR. 
Did you do it with Cyril?

S

jecisc:
I know what is the goal and I'm planning to review it. I just got busy today

Ducasse:
Ok thanks cyril!

jecisc:
Hum... I tried it and this reduces the usability because when we validate it takes when is in the input field and not in the list. 

I'll need to think better about how to improve the UX of this presenter

Ducasse:
OK so let me know I will tag it as work needed.

jecisc:
It took me a while to have a behavior that I found really natural but I think I got it!

I started from your changes Alexis and added other changes and now the flow of this tool seems right!
I'm closing in favor of this PR: https://github.com/pharo-spec/NewTools/pull/983

PR #920, opened 2024-12-13
Breakpoints in debugger

### LINKED ISSUE #5211: Feature request - debugger able to show/set breakpoints
As I am starting to use breakpoints instead of `self halt` I am feeling awkward in the debugger in two aspects:
1. Not being able to see if there is a breakpoint set somewhere in the method being debugged.
2. Not being able to add a breakpoint half way down the method I'm debugging.  

I feel this is a big gap in the breakpoint usability.
How hard would it be to add a left margin similar to the browser to show/set breakpoints?
Who is best to discuss this with further?

StevenCostiou:
Fixes https://github.com/pharo-project/pharo/issues/5211
Thanks to joint work with @adri09070 and @MarcusDenker 

**Right click on the instruction you want to break on:**

Capture d’écran 2024-12-13 à 11 32 44

**The bug icon appear in the gutter, on which clicking opens the dynamic breakpoint configurator:**

Capture d’écran 2024-12-13 à 11 31 30

### DIFF
--- a/src/NewTools-Debugger/DebugPoint.extension.st
+++ b/src/NewTools-Debugger/DebugPoint.extension.st
@@ -1,0 +1,9 @@
+Extension { #name : 'DebugPoint' }
+
+{ #category : '*NewTools-Debugger' }
+DebugPoint class >> allInstalledInMethod: aCompiledMethod [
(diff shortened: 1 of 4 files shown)


StevenCostiou:
Unrelated failing test.

MarcusDenker (approved):

PR #926, opened 2024-12-20
Debug action events small improvements

StevenCostiou:


StevenCostiou:
Information: this PR is still being working on.
It's just slow.

Ducasse:
Ok 
Steven I would like that we pass an application to the model so that we can do 
`(application toolNamed: #browser) open` instead of `Smalltalk tools....`

Basically the presenter should pass its application to the model.

Ducasse:
Do we close this PR since it is from 2024?

PR #927, opened 2024-12-24
COmmenting out Sindarin for now because it blocks a complex PR.

Ducasse:


StevenCostiou:
Any idea why it blocks the other PR?
Can you provide the link of that other PR in this conversation as a trace?

StevenCostiou:
Is that one https://github.com/pharo-project/pharo/pull/17553 ?

StevenCostiou:
By the way that might be the opportunity to rename "ScriptableDebugger" into "Sindarin".
That will be a story for next year.

Ducasse:
Hi steven
This is two days that I'm trying to understand the problem and I have the impression that the code I changed is in Sindarin but the build loads from NewTools. 
So I would like to decompose the problem because once the other PR passes it will be simple to fix.

Ducasse:
Now I get the same problem with https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-17553/25/testReport/MacOSX64.NewTools.Debugger.Breakpoints.Tools/StHaltAndBreakpointControllerTest/osx_64___Tests_osx_64___testDisableHalt/

I renamed the class of the AST and I deprecated them but still the system does not find the new class.

StevenCostiou:
These all look like extension methods of the original AST* classes.
Are they correctly "migrated" as extensions to the new OC* classes by the refactoring?

PR #928, opened 2024-12-25
Rename opal

Ducasse:


Ducasse:
May be I will just redo it. 
Because this is simple.

PR #934, opened 2025-01-12
One less reference to RBParser + does not uselessly reformat code.

Ducasse:
This way opening the inspector will be faster too.


Ducasse:
Broken tests are unrelated so I merge

PR #940, opened 2025-01-16
added WindowManager package to exploit new SpClosedWindowPresenter

AlexisCnockaert:
to handle last closed window


Ducasse:
We should rename it Presenter and add it to the baseline @jecisc Cyril we will need you

PR #943, opened 2025-01-18
Left Over Remembering Size

Ducasse:
The initialExtent: message should be removed.


Ducasse:
More to come because the duplicated logic was not fully cleaned up.

Ducasse:
We will redo it in a more systematic way

jecisc:
Should this PR be merged now?

PR #947, opened 2025-01-19
Branch: 946-StFinder-produces-DNU-on-browse,  Fixes: #964

Ducasse:
But I could not make the application not nil in the caller of StFinderClassSearch.  So I give up.


Ducasse:
I will not work on it anymore. I  published the PR because it should solves the problem but for that we need to know how to correctly configure the object and that I do not know. 
If you do not like it or do not work on it, just close the PR or let it rot.

PR #948, opened 2025-01-19
Branch: FixingLastInitialExtentForWindow

Ducasse:
We pay the price of letting technical debt and wrong API around but the house elves are magically cleaning. :(


MarcusDenker:
Why "Branch:" ? every PR is a branch, or not?

PR #953, opened 2025-01-20
Fix Code Rewriter to take care of the context's application

### LINKED ISSUE #950: Fix StRewriterRuleEditorPresenter buttons to open popups on the same application
I open the presenter with the Toplo backend, and 3 last buttons open a window in the Morphic backend (the default on my image).

Image

Script:
```
app := SpApplication new 
  useBackend: #Toplo;
  yourself.
(StRewriterRuleEditorPresenter newApplication: app) open
```

tinchodias:
Fixes #950


jecisc (approved):

jecisc:
I think that a cleaner step would be to make StChangesBrowserPresenter inherits from StPresenter but that can be done later

tinchodias:
> I think that a cleaner step would be to make StChangesBrowserPresenter inherits from StPresenter but that can be done later

Do you want that I open an issue for that task? (i ignore the consequences of doing such superclass change)

PR #963, opened 2025-01-27
Closed window is now independent and works with Announcements

AlexisCnockaert:
Also cleaned the class and classified

### DIFF
--- a/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
+++ b/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
@@ -41,23 +41,40 @@ SpClosedWindowListPresenter class >> defaultPreferredExtent [
 	^ 310 @ 185
 ]
 
-{ #category : 'example' }
-SpClosedWindowListPresenter class >> exampleClosedWindowsList [
-
-	<sampleInstance>
-	| presenter |
-	presenter := self new
-		             items: SpWindow allInstances;
-		             unhideItemBlock: [ 1 ];
-		             open;
-		             yourself.
-	^ presenter
-]
-
 { #category : 'accessing' }
 SpClosedWindowListPresenter class >> lastClosedWindows [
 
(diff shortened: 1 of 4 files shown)


src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
```diff
@@ -0,0 +1,263 @@
...
+
+{ #category : 'accessing' }
+SpClosedWindowListPresenter class >> lastClosedWindows [
+
+	^ lastClosedWindows ifNil: [  lastClosedWindows := OrderedCollection new ]
+]
+
+{ #category : 'world menu items' }
+SpClosedWindowListPresenter class >> windowMenuOn: aBuilder [
+	"To inject your custom items in a specific place, note that each item increases in order by 1. For example, if you wanted your item to be the second item, you could give it an #order: 1.5"
```
jecisc: This comment could be removed

jecisc (approved):
Don't you need to update the baseline as well?

Ducasse:
BTW this morning I removed some code in SystemWindow to remove the undeclared.
May be we will have to do another pass there.

src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
```diff
@@ -0,0 +1,263 @@
...
+	],
+	#classInstVars : [
+		'lastClosedWindows'
+	],
+	#category : 'NewTools-WindowManager',
+	#package : 'NewTools-WindowManager'
+}
+
+{ #category : 'layout' }
+SpClosedWindowListPresenter class >> defaultLayout [
```
Ducasse: for the default layout it is also nice to move it to the instance side.
Like that we can refer directly to the instance variables instead of symbols.

Ducasse:
Broken tests are unrelated.

PR #966, opened 2025-01-28
Cleaning title definition

Ducasse:
title definition


Ducasse:
I cannot reproduce the failure of testOpenDebuggerWithLastKnownExtent.
In addition I find really strange to have some data stored on a setting class.

Issue #974, opened 2025-01-31
IR inspection of CompiledMethod does not work (ends in error)


JanBliznicenko:
"IR" inspection of `CompiledMethod` does not work (ends in error), because `OCIRMethod` does not have `inpectionIr`. It is caused by #973 (the method is in wrong class)

JanBliznicenko:
Fixed by [#975](https://github.com/pharo-spec/NewTools/pull/975)

PR #975, opened 2025-01-31
Moved IRMethod extension to OCIRMethod

### LINKED ISSUE #973: IRMethod extension should be moved to OCIRMethod
`IRMethod` got renamed to `OCIRMethod` in Pharo 13, but `inspectionIr` extension method (from NewTools) is still in `IRMethod` (its only own method). It probably should be moved.

JanBliznicenko:
`IRMethod` got renamed to `OCIRMethod` in Pharo 13, but `inspectionIr` extension method (from NewTools) was still in `IRMethod` (its only method). It caused `CompiledMethod` "IR" inspection to fail. This PR moves the method. Fixes #973 and #974


Ducasse:
Thanks a lot!

PR #978, opened 2025-02-03
updated the baseline and added a package for tests on SpClosedWindowListPresenter

AlexisCnockaert:
Almost every special behaviour of the presenter is tested I think


AlexisCnockaert:
@Ducasse

PR #979, opened 2025-02-03
Making presenter a singleton

AlexisCnockaert:
-Made SpClosedWindowListPresenter a singleton so it keeps an unique subscription to the closure of window.
-Allowing to switch "modes" between using the logic of unclose window or not
-Added tests on the whole presenter to make sur it works like wanted

### DIFF
--- a/src/BaselineOfNewTools/BaselineOfNewTools.class.st
+++ b/src/BaselineOfNewTools/BaselineOfNewTools.class.st
@@ -264,8 +264,7 @@ BaselineOfNewTools >> postload: loader package: packageSpec [
 
 	MCMethodDefinition initializersEnabled: initializersEnabled.
 	self loadThemeImages.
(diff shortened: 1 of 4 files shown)


jecisc:
There is one failing test

SpClosedWindowListPresenterTest
 ✗ #testSubscriptionToWindowClosedAnnouncements (2ms)
TestFailure: Got 2 instead of 1.
SpClosedWindowListPresenterTest(TestAsserter)>>assert:description:resumable:
SpClosedWindowListPresenterTest(TestAsserter)>>assert:description:
SpClosedWindowListPresenterTest(TestAsserter)>>assert:equals:
SpClosedWindowListPresenterTest>>testSubscriptionToWindowClosedAnnouncements ...assert: presenter items size equals: 1
SpClosedWindowListPresenterTest(TestCase)>>performTest

PR #981, opened 2025-02-04
Move keymapping UI to NewTools

jecisc:
To fix a dependency test


Ducasse:
Hi 

we have
```
SystemDependenciesTest
 ✗ #testExternalSpec2Dependencies (6041ms)
 ✗ #testExternalBasicToolsDependencies (21ms)
 ✗ #testExternalUIDependencies (14ms)
```
Tell me if I should integrate it now and we fix it after

jecisc:
Hum... I'll need to check into more detail what is causing this.

Let's not merge

jecisc:
We should now be able to merge because I'm fixing those tests here: https://github.com/pharo-project/pharo/pull/17770

This change removes unwanted dependencies and we need to remove them on the Pharo side :)

PR #984, opened 2025-02-06
[Suggestion] Making sure the 'save as' treeNavigationPresenter points to the right file system   

Enzo-Demeulenaere:
Hello, this snippet is trying to link the widget file system to the right one, this is useful when working with artificial file systems.

This was a fix done during a sprint to make a new test pass.


Ducasse:
@jecisc 

in case we are looking for. 
```
SystemDependenciesTest
 ✗ #testExternalSpec2Dependencies (13ms)
TestFailure: Given Collections do not match!
	additions : #('NewTools-Core')
	missing: #('NewTools-SpTextPresenterDecorators')
```

jecisc:
I fixed this one last week

PR #985, opened 2025-02-07
Handling shared variables

AlexisCnockaert:
[SystemWindow Shared Variable depends on the SpClosedWindowListPresenter]


AlexisCnockaert:
@jecisc

PR #993, opened 2025-02-11
Moved shortcuts browser from Pharo to NewTools

jordanmontt:
Renamed and moved the shortcuts browser from the pharo repo to the NewToools repo. Now it's called `StShortcutsBrowserPresenter`


jordanmontt:
I saw that this is already done!  https://github.com/pharo-spec/NewTools/pull/981

PR #994, opened 2025-02-11
Pass on the shorcuts browser

jordanmontt:
- Renamed KMDescriptionPresenter to StShortcutsBrowserPresenter.
- Updated the baseline 
- Deleted an old non-loaded "keymap browser"


jecisc (approved):
Thanks!

PR #1005, opened 2025-02-17
Added back unhide with double click and made ux changes

AlexisCnockaert:
- Changed implementation of unhide window when double clicking on the item corresponding
- Updated the title of the presenter to " Window Unclose" since its a tool that we have
- Deleted title since the class already has windowTitle ( according to Spec clean w @Ducasse before )

### DIFF
--- a/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
+++ b/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
@@ -226,6 +226,8 @@ SpClosedWindowListPresenter >> initializePresenters [
 
 	list := self newList
 		        items: OrderedCollection new;
+		        whenActivatedDo: [ :selection |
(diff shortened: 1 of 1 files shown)


jecisc (approved):

jecisc:
There is a failing test:

SpClosedWindowListPresenterTest
 ✗ #testActivateCheckBoxEnableAndDisableSelf (3ms)
#click was sent to nil
UndefinedObject(Object)>>doesNotUnderstand: #click
SpClosedWindowListPresenterTest>>testActivateCheckBoxEnableAndDisableSelf ...click
SpClosedWindowListPresenterTest(TestCase)>>performTest

AlexisCnockaert:
@jecisc like we talked on Friday, we need a new build of Pharo so this test will pass, because its keeping the older Singleton

PR #1006, opened 2025-02-18
Changed the logic behind adding itself in SpClosedWindowListPresenter

AlexisCnockaert:
-Before I checked if the model presenter of the parameter was itself class, but not every window have a model presenter, so test failed
-Now checks the labelString of the parameter, so it don't add "Window Uncloser" in the closed windows


jecisc:
Hum... This PR only adds spaces?

AlexisCnockaert:
Sorry I committed the wrong thing

jecisc (approved):

PR #1008, opened 2025-02-18
Adapted SpClosedWindowListPresenter tests to updates

AlexisCnockaert:
-Since now the presenter is off by default, some tests were failing due to the singleton state
-Now added ``` presenter activateBox click``` calls to make it the same state as enableCloseWindow sharedVariable
-Put the shared variable to true in the setUp in case the Window Unclose is disabled in the current image.
@jecisc this fix the test failures in Pharo normally


jecisc:
IMO you should save the value of the setting in a variable in the setup. Change the value. And set the old value back in the tearDown to be sure the test does not change the value set by the user

PR #1009, opened 2025-02-19
New behavior for the Window Uncloser

AlexisCnockaert:
-After synchronizing we said that we wanted for the presenter to not store undesired windows, like debuggers or inspectors or itself (useless behavior or could cause bugs)
-Added a collection rejectedWindows that has protocols (add/remove/reset) so the user can customise it like he wants to reject certain type of windows

### DIFF
--- a/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
+++ b/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
@@ -30,7 +30,8 @@ Class {
 		'removeButton',
 		'unhideItemBlock',
 		'removeItemBlock',
-		'activateBox'
+		'activateBox',
+		'rejectedWindows'
 	],
 	#classVars : [
 		'UniqueInstance'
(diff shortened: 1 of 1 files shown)


src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
```diff
@@ -264,6 +272,29 @@ SpClosedWindowListPresenter >> initializePresenters [
...
+	rejectedWindows := OrderedCollection new.
+	rejectedWindows add: StDebugger.
+	rejectedWindows add: StInspectorPresenter.
+	rejectedWindows add: self class
+]
+
+{ #category : 'testing' }
+SpClosedWindowListPresenter >> isWindowRejected: aWindow [
+
+	(aWindow model notNil and: [
```
jecisc, line 288: Just a note. We have #notNil and #isNotNil doing the same thing for historical reasons. In general I've seen #isNotNil used more often. 

But I'll still accept the PR, it's just a note

jecisc (approved):

PR #1015, opened 2025-02-25
- Adding object transcript in the baseline

Ducasse:
- Adding registratin to tools
- Better creation API


Ducasse:
Broken tests are unrelated

PR #1021, opened 2025-02-28
add method browser tests

demarey:
add method browser tests and fix browser opening


Ducasse:
The build is strange and probably not related to these changes.

jecisc:
There are two failing tests that seems related

Ducasse:
@demarey I would like to avoid to lose your work.

Ducasse:
The broken tests are unrelated so I will integrate this change.

PR #1023, opened 2025-02-28
Debug point browser fixes

### LINKED ISSUE #1024: Debug point browsing is broken
- removed by https://github.com/pharo-project/pharo/pull/17907
- the original behavior reimplemented "browse" to browse the debug point target (ie asts or methods), which preempted the browsing of the target class from the inspector (original behavior)

StevenCostiou:
Reinsert behavior removed in https://github.com/pharo-project/pharo/pull/17907 to recover browsing behavior.
The behavior is implemented using applications instead of referring directly to Smalltalk tools.
Browsing logic of debug point target fixed: "browsing" browses the target class (as expected) while "browsing target" browses the target entity of the breakpoint (eg the node or object or etc.).

Fixes #1024 https://github.com/pharo-spec/NewTools/issues/1024

### DIFF
--- a/src/NewTools-DebugPointsBrowser/DebugPointBrowserPresenter.class.st
+++ b/src/NewTools-DebugPointsBrowser/DebugPointBrowserPresenter.class.st
@@ -130,7 +130,7 @@ DebugPointBrowserPresenter >> initializePresenters [
 		                   yourself
 ]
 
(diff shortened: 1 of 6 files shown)


StevenCostiou:
Unrelated failings.

PR #1026, opened 2025-03-02
defining a new transcript and its presenter

Ducasse:
This way we will able to kill the old one.


Ducasse:
Unrelated test failing.

PR #1027, opened 2025-03-03
Enhancing object transcript

Ducasse:


Ducasse:
I integrate it and I will fix the problem because I cannot find my image with the PR anymore.

PR #1028, opened 2025-03-05
Fixes #17860 on Pharo

AlexisCnockaert:
- didn't check if a model had a presenter before, causing bugs for the transcript for example
- now checks that the model responds to presenter before checking conditions on it
- Bug fixed :)

### DIFF
--- a/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
+++ b/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
@@ -284,14 +284,13 @@ SpClosedWindowListPresenter >> initializeRejectedWindows [
 { #category : 'testing' }
 SpClosedWindowListPresenter >> isWindowRejected: aWindow [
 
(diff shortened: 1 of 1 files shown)


src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
```diff
@@ -284,14 +284,13 @@ SpClosedWindowListPresenter >> initializeRejectedWindows [
...
-
-
-	^ false
+	^ (aWindow model notNil and: [
+		   (rejectedWindows includes: aWindow model class) or: [
+			   (aWindow model respondsTo: #presenter) and: [
+				   aWindow model presenter notNil and: [
+					   rejectedWindows includes: aWindow model presenter class ] ] ] ])
+		  ifTrue: [ true ]
+		  ifFalse: [ false ]
```
jecisc, line 293: Those two lines are useless. It works without them

PR #1030, opened 2025-03-05
Fix opening of new transcript

jecisc:
Add #open method + update title (the users do not need to know this is a new Transcript) + set the prefered extent to what was in the previous transcript


Ducasse:
Thanks Cyril 
we will have to also set the singleton of the new transcript (model) in the Transcript global variable. The old transcript is installed this way.

I will integrate this PR.

PR #1035, opened 2025-03-07
Make sure that install the transcript does not blow up the system bec…

Ducasse:
…ause morphic could call the stepGlobal method.


Ducasse:
The two failing tests are unrelated and will be fixed by the following PR.

PR #1040, opened 2025-03-12
Introducing CavroisWindowManager, allowing the usage of profiles

AlexisCnockaert:
Pharo13.0.0: #feat: Moving it into NewTools

@Ducasse if you are okay!


Ducasse:
Nice :)

S

PR #1046, opened 2025-03-14
Adding Cavrois to the baseline

Ducasse:


Ducasse:
We should fix the tests of cavrois because 3 are broken since a long time.

Ducasse:
The tests are fixed and the other PR got integrated.

Ducasse:
There is only one dependency test that is not correct

PR #1049, opened 2025-03-14
Refactored and put strings in rejectedPresenters instead of classes

### LINKED ISSUE #1047: Use Sring instead of Class reference
This way we do not have a direct reference to the class because 
this is just a registration mechanims and if the class is not loaded the code should work. 

```
initializeRejectedWindows
	"at the moment we reject only those windows by default"
	
	rejectedWindows := OrderedCollection new.
	rejectedWindows add: StDebugger.
	rejectedWindows add: StInspectorPresenter.
	rejectedWindows add: self class
```

- Rename rejectedWindows -> rejectedWindowClassNames

AlexisCnockaert:
-Did this to prevent code break if some classes that are in are not loaded
-Also made SpClosedWindowListPresenter a subclass of ```StPresenter```
Fixes #1047

### DIFF
--- a/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
+++ b/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
@@ -23,7 +23,7 @@ Public API:
 "
 Class {
 	#name : 'SpClosedWindowListPresenter',
-	#superclass : 'SpPresenter',
+	#superclass : 'StPresenter',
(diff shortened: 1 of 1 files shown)


jecisc:
The problem with strings is that if we rename a class, it will not rename things in this method. 

I think that we could improve by using Symbols instead of Strings. Like this it would not rename them, but at least it would show them in the class references.

PR #1051, opened 2025-03-15
chore: cleaning wrong assumption on dynamic size.

Ducasse:


Ducasse:
I do not get why the tests do not pass since they are green on my machine

PR #1054, opened 2025-03-18
Sindarin improvements and fixes

StevenCostiou:
Just seeing if it breaks anything.
Tests are missing for now.

### DIFF
--- a/src/NewTools-Sindarin-Tools/StSindarinDebuggerPresenter.class.st
+++ b/src/NewTools-Sindarin-Tools/StSindarinDebuggerPresenter.class.st
@@ -38,7 +38,7 @@ StSindarinDebuggerPresenter class >> defaultLayout [
 StSindarinDebuggerPresenter >> buildToolbar [
 	| group |
 	group := CmCommandGroup forSpec.
-	SindarinCommand allSubclasses
+	(SindarinCommand allSubclasses reject:#isAbstract )
 		do: [ :cmd | group register: (cmd forSpecContext: self) ].
 	^ self newToolbar 
 		displayMode: self application toolbarDisplayMode;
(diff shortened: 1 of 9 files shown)


StevenCostiou:
unrelated test failures by the way

src/NewTools-Sindarin-Tools/StSindarinDebuggerPresenter.class.st
```diff
@@ -11,6 +11,12 @@ Class {
 	#package : 'NewTools-Sindarin-Tools'
 }
 
+{ #category : 'accessing' }
```
Ducasse: When you inherit from StPresenter you get automatically the application so you do not need currentApplication
just self application should return it.

StevenCostiou: > When you inherit from StPresenter you get automatically the application so you do not need currentApplication just self application should return it.

Actually it inherits from SpPresenter, which returns SpNullApplication instead of StPharoApplication.
It inherits from it to inherit from StInspectionPresenter.
Looking at the presenter, I don't see the reason why, so it could be migrated to StPresenter hierarchy instead.

jecisc: It's probably historical. IIRC Sindarin started before NewTools was in place

I'd be great to have it inherit from StPresenter if possible

StevenCostiou: I agree, I will see to that tomorrow. Should be simple.

StevenCostiou:
No more failing tests due to my changes.
Can we merge it and move on? :)

jecisc (approved):

PR #1056, opened 2025-03-19
Fix: tests in Cavrois + remove new  from new Transcript

Ducasse:
Fix: tests in Cavrois + remove new  from new Transcript


Ducasse:
Broken test testRestoreUnsavedCodeChanges  is unrelated.

PR #1062, opened 2025-03-26
Debugger better toolbar menu and various fixes/improvements

### LINKED ISSUE #843: Debugger advanced commands should be displayed vertically.
The advanced debugger commands are displayed in a horizontal toolbar.
It might be better to experiment a version in which the actions are displayed vertically, but the Spec toolbar API does not seem easy to tame.

Capture d’écran 2024-09-20 à 09 39 36

### LINKED ISSUE #1061: Debugger config toolbar is greyed when hitting a non-resumable exception
When the context is not resumable, the icons are greyed out as if they were disabled but are still clickable.

They should always be enabled and colored.

Image

StevenCostiou:
Various fixes and improvements:
Fixes #1061 
Fixes #843

The debugger toolbar now has a menu instead of a popup for advanced commands.
Example in mode label+icon (`StPharoSettings toolbarDisplayMode: SpToolbarDisplayMode modeIconAndLabel`) :
Capture d’écran 2025-03-26 à 13 23 25
Capture d’écran 2025-03-26 à 13 23 15
Example in mode icon only (`StPharoSettings toolbarDisplayMode: SpToolbarDisplayMode modeIcon`) :
Capture d’écran 2025-03-26 à 13 23 44

Advanced commands will now be in their own menu if they are defined in an extension group of the debugger toolbar.

Lots of commit because of exploration, better review only changed files rather than commits themselves.


StevenCostiou:
Unrelated failing tests.

PR #1063, opened 2025-03-26
Moving #Profiles menu as a submenu of #Windows

AlexisCnockaert:
Since it's handling windows, it makes sense to put that in the Windows menu
@Ducasse does it make sense to you ? 
For me and Cyril yes


Ducasse:
This is ok for me.

PR #1064, opened 2025-03-26
New debugger extension to have meta-information about the debugging session

StevenCostiou:
This is important for tool building, debugger and interpreter problems/questions exploration, and for some bugs where we want to access information held by the debugger or its session, but not necessarily exposed.

Capture d’écran 2025-03-26 à 17 46 04


StevenCostiou:
Unrelated errors.

PR #1067, opened 2025-03-27
Removed a UIManager dependency

AlexisCnockaert:
-The code was just opening the report of profile action in a TextMorph,  called a Presenter instead to remove dependency
-Fixes #14596 on pharo repo, @Ducasse if you want to check


Ducasse:
Nice :)

Ducasse:
Unrelated failure.

PR #1080, opened 2025-03-28
Fixed the profile update after deletion of current one

AlexisCnockaert:
- The method was looking the wrong thing and was deleting the new current profile it just put before
- Fixed :)


Ducasse:
Ok my gut feeling is that we should write tests.

PR #1082, opened 2025-04-01
Fix cancellation of new protocol creation

### LINKED ISSUE #18054: New protocol is added even though the operation is canceled
When you add a protocol with the "New protocol" menu, if you start typing the name of an existing protocol and you press the Cancel button, the protocol is added anyway while it shouldn't.
This is for Pharo 13.

For example:
- create a new class "MyClass", right click in the protocol column and select "New protocol"

Image

- start typing the name of an existing protocol, like "sort". Notice that the first existing protocol is selected by default

Image

- press the Cancel button. The protocol is added regardless

Image

It might be related this method `StProtocolNameChooserPresenter class>>#requestProtocolNameConfiguring:` and to issue #17978

jecisc:
Fixes https://github.com/pharo-project/pharo/issues/18054

We did not check if the user cancelled until now

### DIFF
--- a/src/NewTools-Core/StProtocolNameChooserPresenter.class.st
+++ b/src/NewTools-Core/StProtocolNameChooserPresenter.class.st
@@ -65,6 +65,8 @@ StProtocolNameChooserPresenter class >> requestProtocolNameConfiguring: aBlock [
 	presenter selectProtocolName.
 
 	dialog open.
+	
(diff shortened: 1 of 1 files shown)


estebanlm:
oups. I am modifying the same part of code. 
I will close this and add your change to my side to avoid conflicts.

PR #1084, opened 2025-04-03
Big changes to make a first good version of Cavrois Profiles

AlexisCnockaert:
- Removed useless presenter class because it is not used here anymore
- Handle multiple windows of the same type by switching back and forth, useful for the user
- Added test so the differents behaviors of the ```CavroisManager```are verified, and the lifetime of profile is detailled
- in ```initialize``` of the manager, we pass the value of the block used in ```ClyBrowserMorph>>open``` to allow browsers to be handled by the manager
- Implemented PopUp strategy so user can choose between a lot of strategies, especially this one can be very useful I think
- Switching profile opens a ```SpConfirmDialog``` to inform the user about the changes and to wait the approbation
- Stack strategy allows now 5 windows max so the workspace is cleaner
- Default save location for a profile is  added to the Manager so it is fully customizable
- Filtered the windows handled so ```SpClosedWindowPresenter``` and ```CavroisWindowManager``` can coexist, + it is useful to keep the window unclose saved in a profile in my opinion
- For now, the big last things to handle before having a clean version for Pharo13 release is:
	- More tests
	- Code cleaning, can be a bit messy in there
	- Details decisions, what do we put at what place so the user experience can be the best possible

@Ducasse tell me if it is good for you !


Ducasse:
sound good now

```
CavroisWindowManagerTest
 ✗ #testConfigurationFromSpSystemWindow (361ms)
 ✗ #testConfigurationFromWindow (76ms)
```
can you check with pablo because may be this is just that the window is not open during the tests

AlexisCnockaert:
it is related to the world and opening problems, I've talked with Pablo.
I think fo r now the best is to remove them (old tests) and rewrite better ones soon

AlexisCnockaert:
And if we have tons of error its because the changes on Pharo are not integrated yet this is normal

Ducasse:
Ok let us retest it.

AlexisCnockaert:
Screen Shot 2025-04-07 at 10 49 39

AlexisCnockaert:
Here is what happens when the user have the same type of window multiple times in a profile ! 

https://github.com/user-attachments/assets/34dafbb2-d175-4c36-a905-c9c612a53b5a

PR #1085, opened 2025-04-04
Adding a Delete Contents button in inspector 

AlexisCnockaert:
To be able to delete files from Pharo, especially when creating a file that a user potentially didn't want to

### DIFF
--- a/src/NewTools-Inspector-Extensions/AbstractFileReference.extension.st
+++ b/src/NewTools-Inspector-Extensions/AbstractFileReference.extension.st
@@ -98,7 +98,7 @@ AbstractFileReference >> inspectionContents: aBuilder [
 			 yourself).
 	toolbarPresenter add: (toolbarPresenter newToolbarButton
 			 icon: (aBuilder application iconNamed: #smallDelete);
-			 label: 'Delete contents';
+			 label: 'Delete file';
 			 help: 'Delete the file';
 			 action: [
 					 self ensureDelete.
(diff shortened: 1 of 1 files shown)


jecisc:
Don't hesitate to add screenshots when you open a PR impacting the UI :)

src/NewTools-Inspector-Extensions/AbstractFileReference.extension.st
```diff
@@ -52,57 +52,69 @@ AbstractFileReference >> inspectionCompressedItemsContext: aContext [
...
+			 help: 'Save contents of file';
+			 action: [
+					 self
+						 ensureDelete;
+						 writeStreamDo: [ :s |
+							 s nextPutAll: contentPresenter text asString ] ];
+			 yourself).
+	toolbarPresenter add: (toolbarPresenter newToolbarButton
+			 icon: (aBuilder application iconNamed: #smallDelete);
+			 label: 'Delete contents';
```
jecisc:
```suggestion
			 label: 'Delete file';
```

"Delete contents" would suggest that it removes the content of the file but not the file itself. I propose to just call if "Delete file"

jecisc (changes requested):

src/NewTools-Inspector-Extensions/AbstractFileReference.extension.st
AlexisCnockaert: Yes it make sense

AlexisCnockaert:
there it is: 

https://github.com/user-attachments/assets/f8512a34-6c71-4614-ab50-b849dd26fc9d

PR #1101, opened 2025-04-14
Updating st object transcript presenter

SebastianLorenzano:


src/NewTools-ObjectTranscript/StObjectTranscriptPresenter.class.st
```diff
@@ -135,12 +135,24 @@ StObjectTranscriptPresenter class >> showInspector: aBoolean [
...
-		parent: #Browse;
+		parent: #InputOutput;
 		label: 'Object Transcript';
-		order: 5;
-		iconName: #playground;
+		order: 3;
+		iconName: #transcript;
+		help: 'Tool to be able to see logs in this instance of Pharo.';
+		action: [ self open ]
+"
```
jecisc, line 148: This should be removed

src/NewTools-ObjectTranscript/StObjectTranscriptPresenter.class.st
```diff
@@ -135,12 +135,24 @@ StObjectTranscriptPresenter class >> showInspector: aBoolean [
...
+
+	<worldMenu>
+	(aBuilder item: #n)
```
jecisc, line 151: Is it normal to have this item name?

SebastianLorenzano: No, it was a typo on my end ty for spotting it

PR #1104, opened 2025-04-16
Implement viewing of MailMessage in Spec

jecisc:
MailMessage has some dependencies over Morphic and UIManager to be able to view a MailMessage content. 

I propose to move those methods to NewTools and to use Spec instead of Morphic to view the content.


MarcusDenker (approved):
ok!

PR #1105, opened 2025-04-22
inspection changes and small fixes

estebanlm:


estebanlm:
this will not work because I restablished the initialPosition to SpWindowPresenter in Spec, but the current image still has the the initialPosition in the wrong place :/ 
But since this are so much changes I cannot be sure so I will do a longer step.

estebanlm:
now this is not related

Issue #921, opened 2024-12-17
StPlayground openUrl: 'http://ws.stfx.eu/6M3DIN9WNU2O' -> DNU


Ducasse:
```
StPlayground openUrl: 'http://ws.stfx.eu/6M3DIN9WNU2O'.
```

Ducasse:
Original issue is #16805

Ducasse:
https://github.com/pharo-project/pharo/issues/16805

jordanmontt:
I will put @astares solution to not forget


> Also it is now StPlaygroundPresenter
We cloud implement a StPlaygroundPresenter(class)>>#openUrl like:
`StPlaygroundPresenter openContents: (ZnEasy get: 'http://ws.stfx.eu/6M3DIN9WNU2O') contents`

SebastianLorenzano:
added in merge request above ☝️☝️

PR #1108, opened 2025-04-24
added StPlaygroundPresenter(class) << openUrl: anUrl

SebastianLorenzano:
(Solution by @astares https://github.com/pharo-spec/NewTools/issues/921)


astares:
@SebastianLorenzano Thanks for sending the PR

PR #1113, opened 2025-04-30
Dynamic code lines show updates in the debugger

StevenCostiou:
- StPharoApplication: listens to settings changes announcements from its application (for now, the "code show line numbers"), and generically redelivers these announcements to its registered presenters
-  StPharoSettings: triggers an announcement when the code show line numbers setting is changed
- StDebugger: dynamically updates its code line numbers on/off when the aforementionned announcement is delivered

Requires https://github.com/pharo-spec/Spec/pull/1765 to be merged first.


StevenCostiou:
Tests cannot pass until https://github.com/pharo-spec/Spec/pull/1765 is in

Ducasse:
What do you think should we push this in P13 or P14?

StevenCostiou:
Depends, I wanted for P13, but perhaps it is too late now or too risky

PR #1116, opened 2025-05-12
revert initialPosition (it is in Spec)

estebanlm:


estebanlm:
yeah, I know this is going to fail. I am integrating it anyway as is the only way to fix the image.

PR #1122, opened 2025-05-15
Go back to old version of ```updateProfile```

AlexisCnockaert:
Made a mistake, old one was better, @Ducasse


Ducasse:
Alexis il faut ecrire des tests sinon on n'est tjrs a la merci de bugs.

PR #1128, opened 2025-05-19
Fix package selection in the new Finder

### LINKED ISSUE #1126: [Finder] Unselected packages cannot be reselected
When selecting only a few packages in Finder, confirming the choice, then reopening the package selection again, those that were not previously selected cannot be re-selected again. They are not shown on the left side of the package chooser window.

Image

jecisc:
Fixes #1126

Now we can select a package we did not select originaly

I also did a refeactoring so that all of the environment search behavior is in StFinderEnvironmentBar and not in StFinderPresenter. This should make the code simpler and not mix the management of the environment in two classes

### DIFF
--- a/src/NewTools-Finder/StFinderModel.class.st
+++ b/src/NewTools-Finder/StFinderModel.class.st
@@ -65,9 +65,9 @@ StFinderModel >> application: anApplication [
 StFinderModel >> availablePackages [
 	"Returns for searching available packages by name."
 
-	^ searchEnvironment packages
-		  collect: [ :package | package name ]
-		  as: OrderedCollection
+	| globalEnv |
+	globalEnv := RBBrowserEnvironment new.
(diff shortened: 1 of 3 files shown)


src/NewTools-Finder/StFinderModel.class.st
```diff
@@ -65,9 +65,9 @@ StFinderModel >> application: anApplication [
 StFinderModel >> availablePackages [
 	"Returns for searching available packages by name."
 
-	^ searchEnvironment packages
-		  collect: [ :package | package name ]
-		  as: OrderedCollection
+	| globalEnv |
+	globalEnv := RBBrowserEnvironment new.
+	^ (globalEnv forPackages: globalEnv packages) packages collect: [ :package | package name ] as: OrderedCollection
```
jecisc, line 70: This is the main change of the PR. Most of the other changes are to move the behavior of environment selection in StFinderEnvironmentBar instead of StFinderPresenter.

Before we looked in a filtered environment, now we look in an environment having all the packages of the image

Issue #1133, opened 2025-05-23
Missing "Method Source" inspection on CompiledMethod


JanBliznicenko:
There is no "Method Source" inspection on CompiledMethod since Pharo13.0-SNAPSHOT.build.643.sha.4790e8fac3.arch.64bit

The following screenshot shows that there is "Method Source" in one build, but not in the other
![Image](https://github.com/user-attachments/assets/b7e2c6e1-cd08-45e5-a57a-c0d1df7628b9)

estebanlm:
crap. That's a regression. I am sure I fixed that last week. 
(and in fact, it was there in the image I used to make the release). 

re-committing...

estebanlm:
merged

JanBliznicenko:
Thank you @estebanlm  for the quick resolution and sorry for the chaotic description in the previous issue :)

PR #1136, opened 2025-05-23
restore CompiledMethod>>inspectionSource:

estebanlm:
restore wrongly removed inspection


estebanlm:
Fixes https://github.com/pharo-spec/NewTools/issues/1133

PR #1145, opened 2025-06-04
Renaming the references to SpWindow to SpWindowMorph

tesonep:
In Spec there was a renaming of the class


MarcusDenker:
Problem: Build now fails with:

```
MetacelloNotification: Loaded -> NewTools-CodeCritiques-Tests --- github://pharo-spec/NewTools:Pharo14/src(545a82f)
Error: Package NewTools-Window-Profiles depends on the following classes:
  SpWindow
You must resolve these dependencies before you will be able to load these definitions: 
```

PR #1147, opened 2025-06-05
fix-and-reintroduce-cpu-watcher

tesonep:
Fix and reintroduce cpu watcher
It depends on https://github.com/pharo-project/pharo/pull/18254


Ducasse:
broken test is unrelated.

PR #1149, opened 2025-06-05
1148 many finder buttons are disabled in pharo13 

pdebruic:
fixes all the buttons when a method or class is selected.


Ducasse:
Thank you. This is great. We are working a lot but help is always welcome!!!!

pdebruic:
I've messed up something in this PR and should not be trying to merge Esteban's commits.  
Not sure how to delete them

PR #1152, opened 2025-06-10
Tiling/start (not finished)

AlexisCnockaert:
Right now we have:

-Menu extension for grids ( remove, cell and grid )
-Two handle drop methods for each mode
-Cell mode is working good, grid mode maybe not as much precise
-Monkey tests for it
-Not linked to profiles yet ( tiling should be used to place windows after)
@Ducasse we can use this to make last changes


Ducasse:
We should pay attention because we have conflicts

Ducasse:
Oops there is a merge conflict.

PR #1153, opened 2025-06-10
Profile menu interaction changes

AlexisCnockaert:
Menu changes inspired by Cyril : 
-Had unclear name for preview functionnality
-Needed a limit for number of profiles
-For better visibility, creation of a submenu for the profile list
-Disable functionalities by default when there's no profile
-Reset windows now handle same window type placement, needed that because it it used too for opening windows


AlexisCnockaert:
@Ducasse those are the changes we talked about + the changes I've talked about with Cyril

Ducasse:
Nice.

Ducasse:
Ok let us merge this.

PR #1155, opened 2025-06-11
Clean breakpoints

guillep:
Newtools has still support for the old breakpoints that are not supported in Pharo since at least Pharo13, superseeded by the DebugPoints.

This PR removes the deprecated support (that looks mostly broken and experimental after some thorough review)


jecisc (approved):

MarcusDenker (commented):
ok!

PR #1164, opened 2025-06-12
Add stpulse

SebastianLorenzano:
Pulse is a remake of Spotter, a front-end to show the result of different processors that provide results. Those processors can be configured in different ways and will provide different access options.

The idea is to divide the old Spotter in 3 tabs:
•  Environment for classes, methods, packages
•  Windows for every open window in the image
•  Tools for menu items

Right now Pulse pulls the first 25 results it gets, getting the first 3 via events so the user doesn't have to wait for the full search to be over to find the first results, which speeds the user experience.
Any time you open either a class, a method or a package (anything that is in the environment tab) it is stored as a HistoryEntry version of itself, which stores either its contents serialized to a string or in the case of the package it stores the package name and when its pulled it makes the search. This was done to evade hard references, and every HistoryEntry are stored in a memory circular logger that cleans itself from any nil entry or any entry with a nil content.
The UI works with a single lists whose context is defined by the page of the paginator that is opened at the moment: When you either change tabs or write, it sends a search (there is a service running every 50 milliseconds which is the one in charge of behaving differently depending on what part of the search is or even if it is or not searching at the moment, StPulse >> processSearch has all the logic and is the method called by the service's step) and there is an outside of service search done when changing tabs.


jecisc:
Hi,

To help with the review it would be nice to add a description of what this PR does :)
I have no idea of what is StPulse

src/NewTools-Pulse/StHistoryClassEntry.class.st
```diff
@@ -0,0 +1,88 @@
+"
+I wrap a Class or Trait entry, and store its content printString, then later using OpalCompiler to compile it into an object when needed.
```
Ducasse, line 2: I do not really get this comment. 
Why do we need to compile it later?
Why Spotter would need that?

src/NewTools-Pulse/StHistoryPackageEntry.class.st
```diff
@@ -0,0 +1,71 @@
+"
+I wrap wrap a RPackage entry name, then later use PackageOrganizer to look for the package (look for self >> content to see the implementation)
```
Ducasse: RPackage does not exist anymore

src/NewTools-Pulse/StPulseModel.class.st
```diff
@@ -0,0 +1,60 @@
+"
+This is the model used in StPulse. 
+
+whenCadidateAddedDo event is to give access to the added candidates to StPulse. See StPulse >> initialize for more info.
```
Ducasse: missing n

src/NewTools-Pulse/StPulseRow.class.st
```diff
@@ -0,0 +1,54 @@
+"
+Row presenter for StPulse list presenter.
+"
+Class {
+	#name : 'StPulseRow',
```
Ducasse: Could it be renamed StPulseRowPresenter since this is a presenter

src/NewTools-Pulse/StHistoryClassEntry.class.st
SebastianLorenzano: It is stored as a string and not as a reference of the object to evade making a hard reference of it. Storing a hard reference would make it so the number of references of the entry never reach 0 and it never goes through garbage collection

src/NewTools-Pulse/StHistoryPackageEntry.class.st
SebastianLorenzano: I meant to say StPackageEntry, went back and fixed it with the typo :)

SebastianLorenzano:
@jecisc You are right, fixed :)

src/NewTools-Pulse/StHistoryClassEntry.class.st
estebanlm: basically, it avoids a leak in case of class/method removal :)

estebanlm:
> @jecisc You are right, fixed :)

but what is StPulse was told

estebanlm:
@SebastianLorenzano errors are relevant to your issue ;)

estebanlm:
errors not related, ready to merge

PR #1165, opened 2025-06-12
Fix for okbutton rename protocol

### LINKED ISSUE #18277: Rename protocol does not work anymore
I tried to rename a protocol but when I select the new name and press the ok button nothing happens.

Pharo version: 14

AlexisCnockaert:
The rename wasnt working using the ok button.
@jecisc fix https://github.com/pharo-project/pharo/issues/18277


jecisc:
This PR breaks a test. I'll fix that

PR #1166, opened 2025-06-12
Improve new transcript

### LINKED ISSUE #1160: StTranscriptPresenter example ends up with an error
StTranscriptPresenter class>>#example ends up in an error. Tested in Pharo 13 and 14 with their default shipped NewTools versions

jecisc:
- Fix example
- Add smoke test on examples
- Allow to have a non global transcript (and rename the singleton accessor from #uniqueInstance to #defaultInstance)

Fixes #1160


Ducasse:
Tx!

PR #1173, opened 2025-06-17
Addition of cellMode to position Window

AlexisCnockaert:
In Cavrois theres now new functionnality so the user can place its windows more precisely using cells
- Initialized morphic blocks in the manager to remove cavrois code from Morphic 
-Added methods to handle cells creation, window drop and drag, world resize 
-New tests related to cells
-Edited the profiles menu to have cells, and linked it to profile creation
-Separated test package to have a clean structure
-Updating the profile preview if an user update its profile/strategies while in preview to avoid an extra step

### DIFF
--- a/src/BaselineOfNewTools/BaselineOfNewTools.class.st
+++ b/src/BaselineOfNewTools/BaselineOfNewTools.class.st
@@ -12,280 +12,270 @@ Class {
 BaselineOfNewTools >> baseline: spec [
 
 	<baseline>
-	spec for: #common do: [ 
-		spec preLoadDoIt: #preload:package:.
-		spec postLoadDoIt: #postload:package:.
+	spec for: #common do: [
+			spec preLoadDoIt: #preload:package:.
+			spec postLoadDoIt: #postload:package:.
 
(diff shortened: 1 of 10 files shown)


AlexisCnockaert:
@Ducasse in addition of the blocks code on pharo

Ducasse:
Strange I integrated the block but so far I get 
````
 > Done in 11883.000ms.
Instance of WorldMorph class did not understand #resizeActionBlock:
WorldMorph class(Object)>>doesNotUnderstand: #resizeActionBlock:
CavroisWindowManager>>initializeBlocks
CavroisWindowManager>>initialize
Cavrois
````

AlexisCnockaert:
its strange, do you integrated the pharo block side in WorldMorph?

AlexisCnockaert:
you need to reset the cavrois manager I think, but when integrated this is done automatically in a new image I guess

Ducasse:
I do not get why the build is failing.
Is the method defined on the class side?

AlexisCnockaert:
its because the pharo new code on WorldMorph is not integrated yet

PR #1175, opened 2025-06-20
Add project-specific script repositories to Sindarin and make 'Create Command' button working

FlavienVolant:
Add project-specific script repositories to Sindarin

https://github.com/user-attachments/assets/0e3e2b05-65e5-4f13-aa24-a66099a35ea0

 'Create Command' button

https://github.com/user-attachments/assets/ead21766-93e6-4067-9c30-76e38f357699

### DIFF
--- a/src/NewTools-Sindarin-Tools/StSindarinDebuggerCreateCommandPresenter.class.st
+++ b/src/NewTools-Sindarin-Tools/StSindarinDebuggerCreateCommandPresenter.class.st
@@ -33,25 +33,31 @@ StSindarinDebuggerCreateCommandPresenter >> connectPresenters [
 
 { #category : 'actions' }
 StSindarinDebuggerCreateCommandPresenter >> createCommand [
-	| className description |
-	className := 'Sindarin', (commandName ifNil: ['Default']), 'Command'.
-	description := descriptionInput text.
+	| name className description class |
+	name := (commandName ifNil: ['Default']) asCamelCase.
+	className := 'Sindarin{1}Command' format: {name}.
+	description := descriptionInput text ifNil: [''].
 	
-	SindarinCommand << className 
+	class := SindarinCommand << className 
 		package: package;
 		build;
 		install.
+
+	class compile: ('execute
+	|sindarin|
+	sindarin := self context sindarinDebugger.
+	self context debuggerActionModel preventUpdatesDuring: [
+		{1}
+	]' format: {'"TODO"'}) classified: 'executing'.
 	
-	"execute"
+	class class compile: ('defaultIconName
+	^#{1}' format: {iconNameSelected}) classified: 'initialization'.
 	
-	"defaultIconName"
-	"iconNameSelected"
+	class class compile: ('defaultDescription
+	^ ''{1}''' format: {description}) classified: 'initialization'.
 	
-	"defaultDescritpion"
-	"description"
-	
-	"defaultName"
-	"nameInput text asCamelCase"
+	class class compile: ('defaultName
+	^ ''{1}''' format: {name}) classified: 'initialization'.
 	
 ]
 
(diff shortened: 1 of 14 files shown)


FlavienVolant:
@StevenCostiou

src/NewTools-Sindarin-Tools/StSindarinDebuggerCreateCommandPresenter.class.st
```diff
@@ -0,0 +1,153 @@
...
+	class class compile: ('defaultName
+	^ ''{1}''' format: {name}) classified: 'initialization'.
+	
+	self createMenu.
+
+]
+
+{ #category : 'actions' }
+StSindarinDebuggerCreateCommandPresenter >> createMenu [
+	StDebugger class compile: ('buildSindarin{2}ExtentionCommandsGroupWith: stDebuggerInstance forRoot: rootCommandGroup
```
jecisc: Is it normal to have this?

If yes, it might be nice to add a comment to explain why we have a method compiling menus :)

src/NewTools-Sindarin-Tools/StSindarinDebuggerCreateCommandPresenter.class.st
```diff
@@ -0,0 +1,153 @@
...
+	
+	toolbarCommandGroup := CmCommandGroup forSpec
+		beToolbarPopoverButton;
+		name: ''{1}'';
+		icon: (stDebuggerInstance application iconNamed: #changeUpdate);
+		yourself.
+	
+	(self debuggerToolbarCommandGroupFor: rootCommandGroup) register: toolbarCommandGroup.
+	
+	commands do: [ :c | toolbarCommandGroup register: c ].' format: {projectName. projectName asCamelCase . package})
```
jecisc:
```suggestion
	commands do: [ :command | toolbarCommandGroup register: command ].' format: {projectName. projectName asCamelCase . package})
```

In Smalltalk we try to avoid abbreviation. Since we have completion tool it's better to have a readable code than to gain 2sec to type the code

src/NewTools-Sindarin-Tools/StSindarinDebuggerScriptingPresenter.class.st
```diff
@@ -35,7 +37,16 @@ StSindarinDebuggerScriptingPresenter >> accept: aVisitor [
 
 { #category : 'actions' }
 StSindarinDebuggerScriptingPresenter >> createCommandFromScript [
-	self flag: 'todo'
+	self flag: 'todo'.
```
jecisc: I see the code was there already, but is the todo still needed? If yes, it'd be nice to add a comment to say what needs to be done

jecisc (commented):
I did a few comments

Also do not hesitate to format methods using cmd/ctrl + shift + f.

It removes for example useless spaces or dots

src/NewTools-Sindarin-Tools/StSindarinDebuggerCreateCommandPresenter.class.st
Ducasse: I do not think that relying on the compile is good. Especially since we will probably change this in Toplo

src/NewTools-Sindarin-Tools/StSindarinDebuggerScriptingPresenter.class.st
FlavienVolant: I created this pr for @StevenCostiou to review my code but it's not finished yet. I left the todo because there is a display issue when I generate a new command group.

src/NewTools-Sindarin-Tools/StSindarinDebuggerCreateCommandPresenter.class.st
FlavienVolant: The goal is to generate a command group for the debugger.  To do that I need to add a class side method to StDebugger. I followed 5.4 "Adding a new menu extending the debugger" from https://github.com/The-Pharo-Debugger-Book-Chapters/Sindarin-Scriptable-Debugger.

If there is a better way to add a method without using compile, I'm open to using it.

StevenCostiou: We need to compile code, so at some point we have to call "compile:". 
I do not understand the relation between compiling code and Toplo...?

I would say though the "createMenu" is not a correct method name.

PR #1176, opened 2025-06-23
display the strategies title instead of class name

AlexisCnockaert:
much better for the user - cleaner preview
@Ducasse should have done that before


jecisc:
Alexis, can you add this method also?

```st
CavWindowStrategy class >> title
	^ self subclassResponsibility
```

AlexisCnockaert:
yes for sure!

PR #1180, opened 2025-06-25
Fixed the resize bug and improved code

AlexisCnockaert:
- Gave kind to the visual morph to avoid a bug when removing one
- Improved ```updateProfile``` before it generated everything and didn't keep old strategies, now just add the extra
- Called build event for window since it wouldn't capture the Spec windows before, causing a bug for the resize


AlexisCnockaert:
don't merge it I will add tests

PR #1182, opened 2025-06-25
Cleaning debugger action model: less dependencies on ast scope

guillep:
Minimise dependencies to the compiler

### DIFF
--- a/src/NewTools-Debugger/StDebuggerContextPredicate.class.st
+++ b/src/NewTools-Debugger/StDebuggerContextPredicate.class.st
@@ -104,16 +104,7 @@ StDebuggerContextPredicate >> printDescription [
 { #category : 'printing' }
 StDebuggerContextPredicate >> printDescriptionPrefixOn: str [
 
-	self printPostMortemDescriptionOn: str.
-	self printHaltDescriptionOn: str
-]
-
-{ #category : 'printing' }
-StDebuggerContextPredicate >> printHaltDescriptionOn: str [
-
-	context sourceNodeExecuted isHaltNode ifFalse: [ ^ self ].
-	str << 'Halt in'.
-	str space
+	self printPostMortemDescriptionOn: str
 ]
 
 { #category : 'printing' }
--- a/src/NewTools-Debugger-Tests/StDebuggerActionModelTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerActionModelTest.class.st
@@ -862,37 +862,6 @@ StDebuggerActionModelTest >> testProceedDebugSession [
 ]
 
 { #category : 'tests - actions' }
-StDebuggerActionModelTest >> testRecompileMethodToInBlockContext [
-
-	| contextChanged previousASTScope |
-	self newSessionWithBlockContext.
-
-	DebugSession contextModelClass: StDebugContextForTests.
-	contextChanged := debugActionModel topContext.
-	
-	"We stepped into a block for which the debugActionModel holds an ast scope:"
-	previousASTScope := debugActionModel previousASTScope.
-	self
-		assert: previousASTScope
-		identicalTo: contextChanged astScope.
-
-	debugActionModel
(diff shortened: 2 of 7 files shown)


src/NewTools-Debugger/StDebuggerContextPredicate.class.st
```diff
@@ -104,16 +104,7 @@ StDebuggerContextPredicate >> printDescription [
...
-	self printPostMortemDescriptionOn: str.
-	self printHaltDescriptionOn: str
-]
-
-{ #category : 'printing' }
-StDebuggerContextPredicate >> printHaltDescriptionOn: str [
-
-	context sourceNodeExecuted isHaltNode ifFalse: [ ^ self ].
-	str << 'Halt in'.
-	str space
```
StevenCostiou, line 116: Why do we remove the halt description?
It could perhaps be done better, but it was there because there was not enough information in the context to crrectly print hatls. 
Also, there seems to be no relation with the purpose of this PR.

StevenCostiou (commented):
Overall, except one change that seem unrelated (the halt print that I would not change), I understand that the dependency is move from what we call the "model" to the debugger presenter, closer to the place where it is needed (and therefore used in only one place):
- It seems better, 
- A test is removed because it was deprecated but no test was added to cover the change,
- There is an unrelated change that should not be there.

guillep:
Thanks for the review!!

Well, there were two points.

1. A *silly thing* is that we could not see where this message was being shown!

We tracked this to the title of the debugger, but we tested since Pharo11, and when a debugger is open on a halt, we have just a `Halt` title.

imagen

So it seems that somehow this code is never called (or the title gets overridden).
Our train of though was kinda "ah, not called often, so is it really worth it?"

2. the call to #sourceNodeExecuted is actually compiler dependent... (it may eventually call the compiler to build the bytecode to AST cache) Maybe we need a higher level API on context -- or the debug info we are working on for what it's worth it -- `isStoppedOnBreakpoint` or something like that? That would allow to have different implementations eventually.

Anyways, let's discuss on how we can minimize these dependencies without breaking features, I'm all for it.

StevenCostiou:
This halt things is for when we hit a halt while stepping, because then the actual exception is lost.
I will have to check, perhaps we have means to do better now because I think we actually added the capture of the exception upon halts hit by the interpreter.

Breaking is fine as long as there are tests ;)

guillep:
Looks like the broken test is not related to this PR:

```
ReleaseTest
 ✗ #testNoEmptyPackages (19ms)
TestFailure: an Array(a Package(NewTools-ObjectCentricBreakpoints)) should have been empty
ReleaseTest(TestAsserter)>>assert:description:resumable:
ReleaseTest(TestAsserter)>>assert:description:
ReleaseTest(TestAsserter)>>assertEmpty:
ReleaseTest>>testNoEmptyPackages ...assertEmpty: violating
ReleaseTest(TestCase)>>performTest
```

The changes in this PR do not touch the package `NewTools-ObjectCentricBreakpoints`

guillep:
> This halt things is for when we hit a halt while stepping, because then the actual exception is lost. I will have to check, perhaps we have means to do better now because I think we actually added the capture of the exception upon halts hit by the interpreter.
> 
> Breaking is fine as long as there are tests ;)

Ah, I see it now.

So, let's consider this code:

```smalltalk
1 halt.
a := 7.
1 halt.
```

If I execute that, the debugger pops and the title says "Halt" and not "Halt in ..."

imagen

Then I step two times, and I have the title

imagen

I wonder why we do not have the same mechanism

StevenCostiou:
> I wonder why we do not have the same mechanism

It's stupid.
The first time you halt you do not have a hand on any exception because: history.
Then the second time, wihtin the debugger, you got your hands on the exception and there the printing goes to the preferred way of having a detailed exception. Or perhaps the other way around.

Anyway, it's bad, and I forgot about it but fixing it should be simple with some late improvements in P13.
I will have a look.

PR #1183, opened 2025-06-26
Small fix cavrois window manager

SebastianLorenzano:
When I pulled a new image of Pharo 13 and Pharo 14 versions of both Spec and NewTools to try out StPulse integration, I noticed that CavroisWindowManager class >> cellModeItem: was getting self current cells as nil when I opened either StPulse or StSpotter. I moved the initialization of the variable from the initialize method to the getter to make sure 100% of the time the getter returns a not nil object


AlexisCnockaert (approved):
Yeah thats a good point

PR #1184, opened 2025-06-27
Fix wrong place for entries

estebanlm:


Ducasse:
Tx it was breaking the build.

Issue #1172, opened 2025-06-16
We cannot do senders, implementors in the new textual transcript


Ducasse:
<img width="886" alt="Image" src="https://github.com/user-attachments/assets/f5b27c74-5751-40f5-a509-4b6945410115" />

Ducasse:
It would be nice to have all the following


<img width="680" alt="Image" src="https://github.com/user-attachments/assets/ab9f63d6-4fe2-4ba0-85cb-9967c50c88d2" />

SebastianLorenzano:
Working on it 👍

PR #1185, opened 2025-06-29
1172 implementors in the new transcript

SebastianLorenzano:
changed text variable from a SpTextPresenter to a SpCodePresenter and removed the syntaxHighlight and lineNumbers. This allows for things like for e.g  <meta> + p, etc.

fixed tests (they weren't always working due needing a wait to have the presenter updated outside of the main thread. This happens because of the structure of the new transcript: The new transcript has a main instance that sends entries via events, and the instances of the TranscriptPresenter recieve the event and update their text using it. Due to this, a wait is needed for the update to be done, and it needs to happen outside of the main thread.

'deleting extra takeKeyboardFocus' is a commit that I mistakenly put here but since it doesn't really matter because it just removes a dead line of code from the StPulse I will leave it here :)

https://github.com/pharo-spec/NewTools/issues/1172

### DIFF
--- a/src/NewTools-Transcript-Tests/StTranscriptPresenterTest.class.st
+++ b/src/NewTools-Transcript-Tests/StTranscriptPresenterTest.class.st
@@ -4,7 +4,7 @@ Class {
 	#instVars : [
 		'presenter',
 		'transcript',
-		'worker'
+		'backendForTest'
 	],
 	#category : 'NewTools-Transcript-Tests',
 	#package : 'NewTools-Transcript-Tests'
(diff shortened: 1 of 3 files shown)


src/NewTools-Transcript-Tests/StTranscriptPresenterTest.class.st
```diff
@@ -32,8 +33,9 @@ StTranscriptPresenterTest >> testAnnouncementIsWritten [
 	[ 
 		window := presenter open.
 		transcript show: 'test'.
-		300 milliSeconds wait.
-		self assert: presenter text text equals: 'test' ]
+		[
```
Ducasse: I do not know if this correct to use this here. 
@guillep what do you think?

SebastianLorenzano: I had no other way to make the tests work without either making the wait async or halting when making the test. Is there any other way to go through the tests? The reason is the one explained in the commit :P

SebastianLorenzano: changed it to use a backendForTest >> waitUntilUIRedrawed instead, like StPlaygroundTest and most others do :)

PR #1192, opened 2025-07-01
Refactor: rename Refactory* classes in Pharo

balsa-sarenac:
We've renamed these classes in Pharo, so they need to be updated here as well.


Ducasse:
tx balsa.

PR #1193, opened 2025-07-02
Fix: window fails to open when message list is nil initially

balsa-sarenac:
While fixing this: https://github.com/pharo-project/pharo/pull/18363
I had issues with this browser. Maybe this is not needed. This is how I used and it failed:

```smalltalk
ReRenameMethodDriver >> browseOverrides

	| overrides |
	overrides := refactoring violations.
	StMessageBrowser
		browse: (overrides collect: [ :ref | ref realClass lookupSelector: newMessage selector ])
		asImplementorsOf: newMessage selector
```


Ducasse:
Tx

PR #1197, opened 2025-07-08
Improving st pulse (fixed defaultKeyboardFocus not working and refactored tests)

SebastianLorenzano:
-fixed defaultKeyboardFocus bug by using event whenOpenedDo: to do self defaultKeyboardFocus takeKeyboardFocus
this forces focus to be on the defaultKeyboardFocus presenter:  It is needed because for some reason, the windows is trying to assign the focus before the DOM is completed, which in turn makes it so the focus is taken after being assigned.
-Added a new test and tearDown Refactored tests  and it didn't seem needed to put wait as it is not asynch.
-Refactored the code a bit and added a getter for listPresenter to be used on testDefaultKeyboardFocus


SebastianLorenzano:
missclicked and added a bunch of things I shouldn't have. Fixing it next commit

SebastianLorenzano:
Moved pull request to https://github.com/pharo-spec/NewTools/pull/1200/

PR #1215, opened 2025-07-25
Fix #1140: [Finder] Feature request: Allow saving edited method code

### LINKED ISSUE #1140: [Finder] Feature request: Allow saving edited method code
When selecting a found method in Finder, you can edit the code in the text area. However, there is no option to save it. It would be nice to be able to make modifications directly from finder.

JMLF:
We set the submit callback when it's needed otherwise we unset it.

We tried changing the submit callback from the code presenter but calling `whenSubmitDo:` multiple times has side effects and raises an exception after the first time. 

Instead what we do is : 

- When a result is selected, the result stores the corresponding callback in itself (each result items has it own rules).
- We have a single submit callback in the code presenter. 
- When the submit callback triggers we delegate  to the result.

### DIFF
--- a/src/NewTools-Finder-Tests/StFinderSaveCodeTest.class.st
+++ b/src/NewTools-Finder-Tests/StFinderSaveCodeTest.class.st
@@ -1,0 +1,159 @@
+Class {
+	#name : 'StFinderSaveCodeTest',
+	#superclass : 'StFinderTest',
+	#instVars : [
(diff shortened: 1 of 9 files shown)


estebanlm (approved):
Sorry for the delay ;)

MarcusDenker:
Note: I change the title to be better for the Friday mail.

PR #1216, opened 2025-07-30
st-pulse-refactoring-list-counter

SebastianLorenzano:
-Made it so now when the selectedIndex is less than 10 and when there is less than 10 objects in the list, it adds a 0 to the selected Index. this was made to evade the layout constantly resizing when the selectedIndex amount of digits changes.
-Fixed a small bug that made it so in some changes the listCounter would not update when changing tabs

### DIFF
--- a/src/NewTools-Pulse/StPulse.class.st
+++ b/src/NewTools-Pulse/StPulse.class.st
@@ -873,7 +873,6 @@ StPulse >> processAddedCandidates [
 StPulse >> processFullAddedCandidates [
 "We candidatesAddedList removeAll to evade it updating the list again after the whole result is added to the list"
 	| candidates |
-	
 	candidatesFullSearchList ifNil: [ ^ self ].
 	candidatesAddedList removeAll.
 	candidates := candidatesFullSearchList.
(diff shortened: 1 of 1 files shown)


src/NewTools-Pulse/StPulse.class.st
```diff
@@ -1025,7 +1027,9 @@ StPulse >> updateResultListWith: aCollection [
 								self historyEntries ]
 						ifFalse: [ aCollection ].
 				emptyPresenter visible: list isEmpty.
-				listPresenter disableActivationDuring: [ listPresenter items: list ].
+				listPresenter disableActivationDuring: [
+					listPresenter items: list.
+					self calculateListCounterPresenter ].
```
estebanlm, line 1032: why here?

SebastianLorenzano: because if I do it before actually setting the result of the list, it calculates it for the result that was before, which was a bit what we were doing before. Because of that and particularly when changing tabs, it happens that sometimes the number doesn't represent the actual number of items the list has

estebanlm: yes, but as calculateListCounterPresenter is defined as this: 
```smalltalk
calculateListCounterPresenter

	listCounterPresenter label: 
		listPresenter selectedIndex asString, 
		'/', 
		listPresenter items size asString
```
I do not understand why it is placed inside the `disableActivationDuring:` and not right after (since the method has no chance of trigger an activation)

estebanlm:
bah, can't be bad anyway

PR #1222, opened 2025-08-07
Add the possibility to add our own bookmarks

jecisc:
This change allows one to easily add their own bookmarks

### DIFF
--- a/src/NewTools-FileBrowser/StFileSystemModel.class.st
+++ b/src/NewTools-FileBrowser/StFileSystemModel.class.st
@@ -23,6 +23,19 @@ Class {
 	#tag : 'UI'
 }
 
+{ #category : 'adding' }
+StFileSystemModel class >> addBookmarkNamed: aString location: aLocation iconName: aName [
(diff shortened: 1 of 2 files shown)


src/NewTools-FileBrowser/StFileSystemModel.class.st
```diff
@@ -23,6 +23,19 @@ Class {
 	#tag : 'UI'
 }
 
+{ #category : 'adding' }
+StFileSystemModel class >> addBookmarkNamed: aString location: aLocation iconName: aName [
+	"Example: `StFileSystemModel addBookmarkNamed: 'pharo' location: FileLocator preferences / 'pharo' iconName: #git`"
+
+	self bookmarks anyOne addBookmark: (StFileBrowserBookmark name: aString location: aLocation icon: (self iconNamed: aName))
```
demarey, line 30: `self bookmarks anyOne` looks strange. Why not adding to `self bookmarks`?

jecisc: This is because bookmarks contains a bookmark group in reality. 

I'm not sur this is really helpful to be able to have multiple bookmark groups but it's the current model

demarey (approved):

PR #1232, opened 2025-08-20
[Backport] Improve folder selection in the File Browser

### LINKED ISSUE #1229: StOpenDirectoryPresenter didn't return current directory when no one selected
For Pharo 14 and 13.

I use StOpenDirectoryPresenter open in modal way.
When I use Open with none selected directory, I got return a empty Array

It's more logic if he return his current folder. When i choose one directory, I double click on it, and i open, so he doesn't count like his selected, and didn't open it.

```
StOpenDirectoryPresenter new openModal
```

jecisc:
The method returning the selected entries of the file browser has some problems.  First the code suggest that we are filtering folders to not return them, but an #and: was used instead of a #or: so this code is not effective. And this method is used also in the #OpenFolder variant of the file browser so the code should not be here. Second, in a traditional file browser, if we allow to select a folder and we press confirm with no folder selected, then we get the current folder. I'm adapting NewTools FileBrowser to act in the same way.

Fixes #1229

### DIFF
--- a/src/NewTools-FileBrowser/StOpenFileOrDirectoryPresenter.class.st
+++ b/src/NewTools-FileBrowser/StOpenFileOrDirectoryPresenter.class.st
@@ -40,13 +40,12 @@ StOpenFileOrDirectoryPresenter >> selectFiles: aCollectionOfFiles [
 
 { #category : 'menu - accessing' }
 StOpenFileOrDirectoryPresenter >> selectedEntries [
-	"Answer a <Collection> of user's selected <FileReference>s "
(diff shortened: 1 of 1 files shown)


estebanlm:
I don't know if this errors are related? Is about cavrois, it shouldn't... 
but since this is a backport I want a confirmation, @jecisc  ;)

jecisc:
Errors are not related. I don't know what is causing them, but this PR has nothing to do with Cavrois

estebanlm:
ok :)

PR #1236, opened 2025-08-30
Fixing  18505: Pulse is not working

Ducasse:
- adding tests
- adding contentString to StEntry (how could it work without it?)
- make history related method follow lazy initialization (else of course you get bugs).

Fixing https://github.com/pharo-project/pharo/issues/18505


estebanlm:
errors not related

MarcusDenker:
I added a bette title for the friday mail

PR #1237, opened 2025-08-30
No DNU anymore when the message browser is open without messages (app…

Ducasse:
…arently some tools do that).

Add a test


estebanlm:
testNoUnusedClassVariablesLeft error seems relevant

Ducasse:
I harvested the changes my current pr

PR #1239, opened 2025-09-12
Fix potential freezing browser

### LINKED ISSUE #18450: File browser is freezing (some long minutes) the image each time it opens
**Bug description**
In Pharo 13, when openning the "File Browser", the Pharo image freezes each time for long minutes.
context : we use multiple disks with a lot of directories on it... 

**Version information:**
 - OS: Windows 11
 - Pharo Version 13.1.0

**Workaround**
The problem seems to be linked to the number of elements of the upper left part "Directories".
The good solutions should be to expand the tree when requested, not when openning the panel.
For my team i manually commented in **StDirectoryTreePresenter**>>initializePresenters the line 
```Smalltalk
children: [ :aClass | aClass subdirectories ];
```
it becomes : 
```Smalltalk
"children: [ :aClass | aClass subdirectories ];"
```

another solution should be to add a setting to disable the StDirectoryTreePresenter ... 
the best should be to build children only when openning the tree... but i don't know how to do it in spec... 

Cheers 
Eric.

AlexisCnockaert:
Due to the **possible complexity** of subdirectories when opening the ```File Browser```,  now it dont expand the tree when its opened.
Screen Shot 2025-09-12 at 14 14 35

Fixes https://github.com/pharo-project/pharo/issues/18450#issuecomment-3164391609

@ELePors is it okay for you that way ?

### DIFF
--- a/src/NewTools-FileBrowser/StFileSystemPresenter.class.st
+++ b/src/NewTools-FileBrowser/StFileSystemPresenter.class.st
@@ -100,25 +100,22 @@ StFileSystemPresenter >> bookmarksTreeTable [
 ]
 
 { #category : 'initialization' }
-StFileSystemPresenter >> connectPresenters [ 
+StFileSystemPresenter >> connectPresenters [
 	"Here we check that the window is displayed. 
 	If it's not, that means this is a first opening so it should preserve the path selected by the user (through openOnLastDirectory preference, or defaultRoot: message send) and not to transmit the directory selection"
 
 	super connectPresenters.
-	
-	treeNavigationSystem
-		transmitDo: [ : selection | 
-			(self isDisplayed and: [ selection isNotNil ])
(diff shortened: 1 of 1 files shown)


Ducasse:
We could have a setting (or a configuration) to be able to have expansion on opening.

jecisc:
Maybe we should do as in the mac Finder or the windows explorer. 

We have a list of the drives but not as a tree. If you click on it, you navigate it in the right. 
The reason they are doing it this way might be to avoid this kind of freezes?

Ducasse:
it would be nice to be able to decide. 
Because it works well on my machine. 
So if people have slow discs or slow network. 
So we could have a classVar to control the opening because opening on your folder is quite nice.

Ducasse:
I will integrate it now and I will fix it.

PR #1241, opened 2025-09-15
Renaming StDebuggerActionModel to StDebuggerClientModel

### LINKED ISSUE #1070: Debugger presenters and models are wrongly named
In the system, we have the actual debugger model:
- `Context`
- `DebugSession`
- `DebugContext`

But in NewTools we have:
- `StDebugger` that should be named `StDebuggerPresenter`
- `StDebuggerActionModel` that should be named `StDebuggerClient` 
- All associated classes to `StDebuggerActionModel` that uses the same base name should be renamed to `Client`

StevenCostiou:
Fixes #1070


Ducasse:
Hi @StevenCostiou let us know when we merge this change. Now there is conflict and I do not really know what to do. It should be simple.

Ducasse:
I made a mistake when trying to solve the conflict. So all the tests break. So I will revert it.

PR #1251, opened 2025-09-26
adding copy systen info into clipboard under debugger right click menu

### LINKED ISSUE #549: [Debugger][Enhancement] Propose a shortcut to copy system infos to report a bug
It should be very cool to have a shortcut to copy in the system clipboard formatted informations about the system. Theses informations can be used to create an issue (and don't forget to include them).

I create an issue when a debugger is open, so why not to include an item in the debugger menu for that.

image

For example, it can be in **Bug report > Copy system info into clipboard**

The result,, when I do a Ctrl+V, should be : 

```
System config
Pharo : Pharo-11.0.0+build.692.sha.0c234f2f5a92b237dd22ef9429506e92e586c494 (64 Bit)
OS : Windows 11 Professionnel (22H2 Build 22621.1778) x64
```

And why no include others action into **Bug report** menu as a shortcut which open github on a new issue :)

fouziray:
fix #549  by adding a new command in the right-click menu in a debugger window. 
@Julesboul  @DurieuxPol

### DIFF
--- a/src/NewTools-Sindarin-Commands/SindarinCopyOsInfo.class.st
+++ b/src/NewTools-Sindarin-Commands/SindarinCopyOsInfo.class.st
@@ -1,0 +1,41 @@
+Class {
+	#name : 'SindarinCopyOsInfo',
+	#superclass : 'SindarinCommand',
+	#category : 'NewTools-Sindarin-Commands',
+	#package : 'NewTools-Sindarin-Commands'
+}
+
+{ #category : 'default' }
+SindarinCopyOsInfo class >> defaultDescription [
+
+ ^ 'Copy to clipboard system information and Pharo image version'
+]
+
+{ #category : 'initialization' }
+SindarinCopyOsInfo class >> defaultIconName [
+	^#smallCopy
+]
+
+{ #category : 'default' }
+SindarinCopyOsInfo class >> defaultName [
+<toolbarExtensionDebugCommand: 50>
+<codeExtensionDebugCommand:1 >
+	^ 'Copy system information'
+]
+
+{ #category : 'executing' }
+SindarinCopyOsInfo >> execute [ 
+
+Clipboard clipboardText: self systemInfo.
+]
+
+{ #category : 'as yet unclassified' }
+SindarinCopyOsInfo >> systemInfo [
+
+	^ Smalltalk os platformName,
+	 Smalltalk os version ,
+	 Character space asString ,
+	 Smalltalk os subtype ,
+	 Character cr asString,
+	 Smalltalk image lastUpdateString
+]


src/NewTools-Sindarin-Commands/SindarinCopyOsInfo.class.st
```diff
@@ -0,0 +1,41 @@
+Class {
+	#name : 'SindarinCopyOsInfo',
```
jecisc, line 2: Is it really specific to Sindarin to copy some infos?

Seems like something more general than Sindarin

src/NewTools-Sindarin-Commands/SindarinCopyOsInfo.class.st
```diff
@@ -0,0 +1,41 @@
...
+	^ 'Copy system information'
+]
+
+{ #category : 'executing' }
+SindarinCopyOsInfo >> execute [ 
+
+Clipboard clipboardText: self systemInfo.
+]
+
+{ #category : 'as yet unclassified' }
```
Ducasse, line 32: No as yet unclassified

Ducasse:
@fouziray @Julesboul can you reply to @jecisc and do the next step?

src/NewTools-Sindarin-Commands/SindarinCopyOsInfo.class.st
fouziray: I'm not sure we want to have it on all menus. 
According to the issue we want to be able to report system info from a debugger window when an error is raised, to help documenting a new issue about it. 

This way we avoid carrying an extra entry in the right click menu elsewhere where it isn't used as much as on a debugger.

StevenCostiou: I agree with you Faouzi but what Cyril means is that it maybe should be at the "general debugger" level rather than on Sindarin.
That's another subject but in my opinion Sindarin should be removed from the system, and be an external, optional project. If that ever happens, should we still want to have the feature you propose available in the debugger?
If that is so, then it just means you have to rename and move your command class in the debugger packages with other commands ;)

fouziray: Ah now i see what you mean, in this case i'll try to move the command into the debugger. Thanks for clearing this up for me :)

Ducasse: Steven about sindarin, let us know. I liked the idea of being able to script the debugger. 
What I can tell you is that with debug info we paid attention not to break it. 
If it is out of Pharo, it will be probably difficult to maintain it. Now I think that for debug info we will not do more changes. so may be we could remove it when the debugger can use offlinedebuginfo like that it will be in a good shape for the future.

Ducasse:
About this menu item. I have a question what is the different with the information from the About. 
Because may be we could enhance the About (with potential missing information). and avoid to it in the debugger menu.

PR #1254, opened 2025-09-26
filtering protocols when creating a new method in debugger

### LINKED ISSUE #1207: [Debugger] Method creation on class side have a lot of protocols
It seems that when we create a method in the debugger, we get proposed all the protocols up to Object in the hierarchy. This is super nice to have some basic protocols without having too much proposed!

But on the class side we have all the protocols of the class hierarchy. Maybe the filter should be put at a different level in the hierarchy?

fouziray:
Restrained added protocol names when creating a new class-side method in the debugger.

fix #1207
@Julesboul @DurieuxPol

### DIFF
--- a/src/NewTools-Debugger/StDebuggerMethodImplementorPresenter.class.st
+++ b/src/NewTools-Debugger/StDebuggerMethodImplementorPresenter.class.st
@@ -54,17 +54,20 @@ StDebuggerMethodImplementorPresenter >> addProtocolsFrom: aClassOrTrait [
 	aClassOrTrait ifNil: [ ^ self ].
 
 	allProtocolsWithDuplicates := Dictionary new.
+	
 	(aClassOrTrait isTrait
 		 ifTrue: [ aClassOrTrait protocols ]
 		 ifFalse: [
-				 (aClassOrTrait withAllSuperclasses copyWithoutAll: {
+			((aClassOrTrait withAllSuperclasses copyWithoutAll: ({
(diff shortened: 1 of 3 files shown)


src/NewTools-Debugger/StDebuggerMethodImplementorPresenter.class.st
```diff
@@ -57,9 +57,14 @@ StDebuggerMethodImplementorPresenter >> addProtocolsFrom: aClassOrTrait [
...
-					  Object.
-					  ProtoObject }) flatCollect: #protocols ]) do: [ :protocol |
+				 (aClassOrTrait withAllSuperclasses copyWithoutAll: {
+						  Object.
+						  ProtoObject.
+						  Class.
+						  Behavior.
+						  ClassDescription.
+						  Object class.
+						  ProtoObject class }) flatCollect: #protocols ]) do: [ :protocol |
```
jecisc: If the class of trait is directly one of those class, we should not filter its protocols.

IMO we should have the protocols of `aClassOrTrait` and of all its superclasses that are not in this list. 

The problem is that is the class is one of those, we will have nothing suggested at all

jecisc (changes requested):
If possible it would be great to add tests on this

Ducasse:
@fouziray and others. This is important to check your PR and update them.

jecisc:
Failing test does not seems related

PR #1255, opened 2025-09-30
[P12] Fix Issue 16827: Method version history timestamp always shows a wired date

tesonep:
Using the stamp instead of the timestamp. So we avoid trying to parse the string that may content author and different date format.

Fix issue https://github.com/pharo-project/pharo/issues/16827


jecisc (approved):

jecisc:
Should I port it forward P13 and P14?

tesonep:
It is not affecting P14 and P13 because the stamps are correctly generated
and they don't handle author and different formats, just a nice ISO format

Pablo Tesone.
***@***.***

PR #1259, opened 2025-10-09
Using DebugInfo

guillep:
This PR makes use of the new debug info interface.
This API, which is to be used in tools such as debugger, inspector, is responsible of reading variables from a context, and to obtain bytecode to source code mappings:
  - is the current PC a send?
  - is the current PC a store?
  ...

Particular care is taken for optimized/inlined contexts, and avoiding creeping implementation details to users!!

Depends on: https://github.com/pharo-project/pharo/pull/18644


estebanlm:
hey @StevenCostiou, I imaging this is ok, but can you take a look? This is your domain after all :P

StevenCostiou:
I know its there I need to have a look :p
More precisely, we agreed that it would make some tests fail that I will comment or remove until I can fix them.

Ducasse:
So I can merge and we check because this is important that this PR does not rot. 
We found some glitches and need to understand the problem (mainly not with the static version but with the old opal way that we may have shaken a bit ).

PR #1262, opened 2025-10-13
Migrate Epicea browsers to NewTools

jecisc:
A also updated the presenters to inherit from StPresenter and made sure they had the Pharo application


jecisc:
Closing because changes are happening in Epicea and this is hard to sync

PR #1266, opened 2025-10-26
Fixing some CmCommandAborted should be pass an application!!!!!

Ducasse:


Ducasse:
It is dependent on a PR in Spec that depends on a PR in Pharo :(

Ducasse:
The failing tests are unrelated.

PR #1267, opened 2025-10-27
add an accessor for application in morphs 

estebanlm:
for the case we use them in the context of a NewTool (like, when using calypso that is mixed with spec tools things).


estebanlm:
errors unrelated

PR #1268, opened 2025-10-27
move wrong class-side methods to instance-side 

estebanlm:


estebanlm:
errors still not related

PR #1271, opened 2025-10-29
Cut dependency of Monticello to UIManager

jecisc:
Monticello is using the UIManager to ask the user if they want to load/merge or cancel the loading of a package if it would impact dirty packages. This change propose to use Spec to ask this question and cut one more dependency over the UIManager.


Ducasse:
broken tests are unrelated

PR #1277, opened 2025-11-05
Inspecting a string should display the ' ' (Fix #814)

### LINKED ISSUE #814: Inspecting a string should display the ' '
Inspect `x := '#north' asSymbol `

Capture 2024-07-27 at 22 37 18

Julesboul:
Change methods ```Symbol>>#stDisplayString``` and ```String>>#inspectionString:```.
Add methods ```String>>#previewString``` and ```Symbol>>#previewString```.

It answer the #814 issue by printing the '' when printing a symbol with a string.

### DIFF
--- a/src/NewTools-Inspector-Extensions/String.extension.st
+++ b/src/NewTools-Inspector-Extensions/String.extension.st
@@ -25,9 +25,15 @@ String >> inspectionFullString: aBuilder [
 
 { #category : '*NewTools-Inspector-Extensions' }
 String >> inspectionString: aBuilder [
+
 	<inspectorPresentationOrder: -10 title: 'Preview'>
-	
 	^ aBuilder newText
-		text: (self truncateWithElipsisTo: 1000);
-		yourself
+		  text: self previewString;
+		  yourself
+]
+
+{ #category : '*NewTools-Inspector-Extensions' }
+String >> previewString [
+
+	^ self truncateWithElipsisTo: 1000
 ]
--- a/src/NewTools-Inspector-Extensions/String.extension.st
+++ b/src/NewTools-Inspector-Extensions/String.extension.st
@@ -28,6 +28,6 @@ String >> inspectionString: aBuilder [
 
 	<inspectorPresentationOrder: -10 title: 'Preview'>
 	^ aBuilder newText
-		  text: self printString ;
+		  text: (self truncateWithElipsisTo: 1000) printString;
 		  yourself
 ]
--- a/src/NewTools-Inspector-Extensions/String.extension.st
+++ b/src/NewTools-Inspector-Extensions/String.extension.st
@@ -28,12 +28,6 @@ String >> inspectionString: aBuilder [
 
 	<inspectorPresentationOrder: -10 title: 'Preview'>
 	^ aBuilder newText
-		  text: self previewString;
+		  text: self printString ;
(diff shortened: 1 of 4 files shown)


jecisc (commented):
I am really not sure about what I will say but maybe we could use #displayString and override #displayStringOn: on Symbol to add the `#`?

This would be to avoid adding yet another printing method

estebanlm:
I have to agree with @jecisc here. Let's discuss this (thursday, since I will not be available until then ;) )

Julesboul:
Talking with @Ducasse about the wanted fix, I did some changes. 
The idea is to always have the '' in the preview (even for String objects).

So there is no need for adding a new method. Just changing `String>>#inspectionString:` and `Symbol>>#stDisplayString` methods.

However, String are still display without '' in the `StRawInspectionPresenter` and I don't know if we want this to be changed.

src/NewTools-Inspector-Extensions/String.extension.st
```diff
@@ -25,9 +25,9 @@ String >> inspectionFullString: aBuilder [
 
 { #category : '*NewTools-Inspector-Extensions' }
 String >> inspectionString: aBuilder [
+
 	<inspectorPresentationOrder: -10 title: 'Preview'>
-	
 	^ aBuilder newText
-		text: (self truncateWithElipsisTo: 1000);
-		yourself
+		  text: (self truncateWithElipsisTo: 1000) printString;
```
jecisc:
```suggestion
		  text: (self printStringLimitedTo: 1000);
```

src/NewTools-Inspector-Extensions/Symbol.extension.st
```diff
@@ -3,5 +3,5 @@ Extension { #name : 'Symbol' }
 { #category : '*NewTools-Inspector-Extensions' }
 Symbol >> stDisplayString [
 
-	^ '#{1}' format: { super stDisplayString }
+	^ (self truncateWithElipsisTo: 1000) printString
```
jecisc:
```suggestion
	^ self printStringLimitedTo: 1000
```

src/NewTools-Inspector-Extensions/StringTest.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : 'StringTest' }
+
+{ #category : '*NewTools-Inspector-Extensions' }
+StringTest >> testStDisplayString [
```
jecisc: Since you are truncating the strings maybe it might be nice to have a test with a big string?

Little helper for such tests: `String loremIpsum: 1001`

jecisc (commented):
I think we can use printStringLimitedTo: instead of printString to avoid 2 trunctation. I also suggested a new test to add

Julesboul:
> I think we can use printStringLimitedTo: instead of printString to avoid 2 trunctation. I also suggested a new test to add

Thanks, I did the expected changes and add a test for large strings/symbols

jecisc (approved):
Thanks!

Ducasse:
This is strange because in P14 I still have 
Capture 2026-01-27 at 15 13 54

Julesboul:
I just add a pull request about this [here](https://github.com/pharo-spec/NewTools/pull/1331)

PR #1283, opened 2025-11-17
Change preview on string inspection into a code presenter

Julesboul:
This pull request answer [pharo's issue 18914](https://github.com/pharo-project/pharo/issues/18914).

To allow commands like "senders of it", a CodePresenter is required. However, "preview" tab on the Inspector (for Strings and Symbols) was a TextPresenter allowing just "copy", "paste", ...

So the presenter was changed to have this feature.


Julesboul:
Mistake on fork branches, new PR : https://github.com/pharo-spec/NewTools/pull/1286

PR #1286, opened 2025-11-21
Fix/18914

Julesboul:
This pull request answer https://github.com/pharo-project/pharo/issues/18914.

To allow commands like "senders of it", a CodePresenter is required. However, "preview" tab on the Inspector (for Strings and Symbols) was a TextPresenter allowing just "copy", "paste", ...

So the presenter was changed to have this feature.

### DIFF
--- a/src/NewTools-Inspector-Extensions/String.extension.st
+++ b/src/NewTools-Inspector-Extensions/String.extension.st
@@ -29,5 +29,6 @@ String >> inspectionString: aBuilder [
 	<inspectorPresentationOrder: -10 title: 'Preview'>
 	^ aBuilder newCode 
 		  text: (self printStringLimitedTo: 1000);
+		  beNotEditable;
 		  yourself
 ]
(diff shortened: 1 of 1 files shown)


src/NewTools-Inspector-Extensions/String.extension.st
```diff
@@ -27,7 +27,7 @@ String >> inspectionFullString: aBuilder [
 String >> inspectionString: aBuilder [
 
 	<inspectorPresentationOrder: -10 title: 'Preview'>
-	^ aBuilder newText
+	^ aBuilder newCode 
 		  text: (self printStringLimitedTo: 1000);
 		  yourself
```
jecisc, line 33:
```suggestion
		  text: (self printStringLimitedTo: 1000);
		  beNotEditable;
		  yourself
```

We should not be able to edit this pane. This still allows to browse info but we cannot modify the test with this line

jecisc (changes requested):

jecisc:
Also just a little info: If you write `Fixes #issueNumber` or `Fixes {link to issue in other repo}` then the issue get automatically closed when we merge the PR :)

PR #1290, opened 2025-11-24
Fix bug when choosing a protocol name

Julesboul:
This pull request is fixing a bug on `StProtocolNameChooserPresenter`.

Indeed, the first protocol in the suggestion list was always selected and put as the choosen protocol (even if you've wrote something in the text editor).

So now the text editor is chosen by default (you can still select a name in the filtering list).

By the way, it will be very usefull to also add this fix in P13.

### DIFF
--- a/src/NewTools-Core-Tests/StProtocolNameChooserPresenterTest.class.st
+++ b/src/NewTools-Core-Tests/StProtocolNameChooserPresenterTest.class.st
@@ -18,16 +18,6 @@ StProtocolNameChooserPresenterTest >> setUp [
 	presenter := StProtocolNameChooserPresenter new
 ]
 
-{ #category : 'running' }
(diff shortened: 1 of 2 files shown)


jecisc:
If we do this then we should remove the "+" button.

The idea was that most of the time we are using an existing protocol, so we select the first of the list by default and we have the arrow up/down navigation binded. And if we want to create a new protocol we have the "+" button. 

But if the list elements are not selected by default, the plus becomes useless

Julesboul:
Yes indeed! 
But I didn't see any difference between '+' button and 'OK' on the current implimentation : if I'm writting a new protocol name and doing '+' it selects the selected protocol in the list (instead of creating a new one).

jecisc:
There was a difference in the past but other people worked on this presenter since I created it so maybe the behavior I intended to have got lost :(

PR #1294, opened 2025-11-30
use the StMessageBrowser by default

Ducasse:


Ducasse:
The problem was in a previous conflict resolution.

PR #1297, opened 2025-12-02
debugger: using new easy list view for stack view

### LINKED ISSUE #1296: debugger stack is not resizing correctly when few rows
if there is a list inside a box layout which is contained in a paned layout, and the original list has less rows than the needed to overflow the height (which will show a scrollbar) then the list will not receive a scrollbar and no matter how many times I resize the splitter, it will not be resized and receive a consequent scrollbar.

this code reproduces the error : 

```smalltalk
app := SpApplication new.
presenter := SpPresenter newApplication: app.
 list := presenter newEasyListView.
presenter layout: (SpPanedLayout newVertical
    add: (SpBoxLayout newVertical 
        add: list;
        yourself);
    add: (presenter newLabel)
    yourself).
    
list items: (Smalltalk allClassesAndTraits first: 5).
    
presenter open.
```

yes, this is spec and not the debugger, but is here where is visible :)

estebanlm:
- gives me more control (and fixes a long time bug on the stack list when not many rows were shown)
- adding a style contributor to handle styles in a better way than hardcoding in the wrong place.
- using direct presenters in layouts (instead of selectors) as the old way is no longer necesary and is confusing.

fixes https://github.com/pharo-spec/NewTools/issues/1296

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -527,7 +527,7 @@ StDebugger >> codeLayout [
 
 	^ SpBoxLayout newTopToBottom
 		  add: self toolbarPresenter expand: false;
-		  add: #code;
+		  add: code;
 		  yourself
 ]
 
@@ -642,7 +642,7 @@ StDebugger >> defaultLayout [
 	^ SpPanedLayout newTopToBottom
 		positionOfSlider: 65 percent;
 		add: stackAndCodeContainer;
-		add: #inspector;
+		add: inspector;
 		yourself
 ]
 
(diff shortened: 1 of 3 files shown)


StevenCostiou (approved):

StevenCostiou:
Failing test because of a halt in a debug context method? Why is there a halt there? This is bad... really I don't get how the system ends up like that...

Ducasse:
Steven instead of ranting, try to understand that we are chasing all the references to morphic so I probably put an halt to see if the code was used. You have only a little idea how much effort it cost us. 
This is HUGE. but we do it with great happiness and fighting spirit because we will have a better system. 
Notice also that Pharo 14 is under development and spotting an halt for the release is nothing but it helps us to move forward. And I do not have the luxury to only work on the code I know. This is why I also put halt in Object methods and cleaned already a lot of bad dependencies.

Ducasse:
The message I wrote is somehow clear. :)
Consider that UI manager will just vanish.

```
confirm: message trueChoice: trueChoice falseChoice: falseChoice
	self halt.
	"YOU SHOULD NOT CALL UIMANAGER"
	^ UIManager default
		confirm: message
		trueChoice: trueChoice
		falseChoice: falseChoice
```

Ducasse:
Now the related point is that I cannot fix everything magically, manage a team, do the budget and the rest.
So we started to clean inform:, and now we start to clean confirm: and then uiManager, and probably some other after. And confirm:trueChoice:falseChoice: is quite ugly and far on the list.

Ducasse:
Tx esteban this is good to see the code of easylist user since it looks much simpler.

StevenCostiou:
Stéphane, you also rant when you don't like something and when that something is a pain in your path.
Consider me your equal :)

PR #1306, opened 2025-12-22
Clicking on "Close" is closing the window

Julesboul:
When opening the "Window menu" of windows and clicking on "Close", windows were hidden.

So we fixed this to match the expected behavior (i.e. closing the window).

### DIFF
--- a/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
+++ b/src/NewTools-WindowManager/SpClosedWindowListPresenter.class.st
@@ -136,7 +136,7 @@ SpClosedWindowListPresenter class >> windowMenuOn: aBuilder [
 	(aBuilder item: #Close)
 		order: 1.0;
 		action: [
-			aBuilder model hide.
+			aBuilder model close.
 			announcement := WindowClosed new
 				                window: aBuilder model;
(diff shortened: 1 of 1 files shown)


Ducasse:
I think that this is normal that there are hidden and the bug is that hidden windows should not be listed in the taskbar.

AlexisCnockaert:
@Ducasse we did a `Close` option to hide the window and a `Destroy` one to close them for real months ago, do you think we need to change something?

Ducasse:
For now I have the impression that we should not move :)
We have better things to improve.

AlexisCnockaert:
Ok so we can close this!

guillep:
Hi, I come from the outside :) I read the comments and I don't get what's the action to take.

@AlexisCnockaert , @Julesboul , @Ducasse would you mind making a quick sync and tell me if this should be integrated, closed or if there is something else?

Ducasse:
I do not think that we should integrate it but alexis should confirm this. 
The close is a hide in fact.

AlexisCnockaert:
Yes I confirm we should close this ( I can't).
But I think the Destroy option is not very clear because it is the one that closes a window (not hiding it)

PR #1318, opened 2026-01-07
renamed StFileSystemModel into StFileSystemModelPresenter

Ducasse:


jecisc:
That makes sense but some people are using it so we should add a deprecated alias

You can either add this class method:

```
initialize
	self deprecatedAliases: #( StFileSystemModel )
```

Or I can add it later when I'll get the time

Ducasse:
In fact I do not understand why this class inherits from Presenter.
I feel wrong.

PR #1323, opened 2026-01-15
fix: show the finder search options again #19105

gbosetti:
Issue https://github.com/pharo-project/pharo/issues/19105

The menu is back:

Screenshot 2026-01-15 alle 20 02 19


jecisc (approved):
Nice, thank you :)

PR #1324, opened 2026-01-19
Trim the name on protocol creation

### LINKED ISSUE #19178: Protocol names should be trimBoth
Right now we have a boring that reports that protocol names should contain leading or finishing space. I would be 1000 times better that by construction this is not possible to happen. 
Then we can remove the rule.

AlexisCnockaert:
To avoid getting a warning from the rule that looks for spaces in protocol name, I made it impossible by construction trimming the new protocol name. I also added a new test for this behaviour
Thanks to those changes, the rule can be safely removed. ( in Pharo repository, other PR).
Fixes [#19178](https://github.com/pharo-project/pharo/issues/19178)

### DIFF
--- a/src/NewTools-Core-Tests/StProtocolNameChooserPresenterTest.class.st
+++ b/src/NewTools-Core-Tests/StProtocolNameChooserPresenterTest.class.st
@@ -46,6 +46,19 @@ StProtocolNameChooserPresenterTest >> testCreatingProtocolThatExists [
 ]
 
 { #category : 'tests' }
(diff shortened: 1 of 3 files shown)


jecisc:
I think that managing the trim in ClassDescription>>#addProtocol: would be a better place. It would manage also code coming from Tonel, FileTree, metaprogrammation, etc

AlexisCnockaert:
You are right, I will close this one and do all in the pharo one

PR #1325, opened 2026-01-19
Add the default action for the ConfirmationRequest

jecisc:
This needs https://github.com/pharo-project/pharo/pull/19184 to be integrated


Ducasse:
In fact instead of an extension the ConfirmRequest could expose a block setter.

PR #1326, opened 2026-01-20
Speed up the Spotter

jecisc:
I have other ideas to speed up the spotter but here is a first batch.

- Remove StEntry>>#initialize because it does not do anything but it slow down the creation of objects that are instantiated all the time by the Spotter (at least once for each methods, class and package of the system)
- Use #beginsWith:caseSensitive: and #includesSubstring:caseSensitive instead of using #asLowercase everywhere. This is much faster

With those two improvements the tests of spotter went from 17sec to 11sec to run on my machine which is a nice improvement


jecisc:
In fact the failing tests are coming from a bug in Pharo. I'll propose a PR

jecisc:
Relaunching since Pharo's PR got integrated

PR #1327, opened 2026-01-22
new settings browser re-work

estebanlm:
and also replace the fetch all button of Pulse with a link that explicitly says "Fetch all" (button was really unclear).
NOTE: I still need to make a pass for some things... but this is looking better :P


guillep:
Would you mind using clearer commit messages than revert reverted? :)

estebanlm:
error not related

PR #1330, opened 2026-01-24
Fix #1311 with a CleaTranscript announcement (incl. Test)

### LINKED ISSUE #1311: Transcript>>clear doesnt seem to work anymore in P13 or P14
However, it does work in P12...

Presumably this was overlooked when the Transcript was rewritten?

It could have a test case too...

astares:


estebanlm (commented):
LGTM

estebanlm:
I wonder how testUnselectWhenChangingToAllLogsTab can have a DNU there?
in any case, this is not related to this PR

PR #1331, opened 2026-01-28
Fix String display in Inspector

Julesboul:
This pull request fix the display of String objects in the Inspector by adding quotes.

Here is an example of an Array inspection: 

Example_DisplayString_Inspector


Ducasse:
I love it!

PR #1337, opened 2026-01-30
fix issue #1336 by adding a scrollable layout

### LINKED ISSUE #1336: StOpenFilePresenter cancel and open buttons disappear after a certain directory depth
When searching for a file nested deeply enough in directories, the sidebar on the left showing the file tree will take more space. This can lead to the modal buttons ("Close" and "Open") to be pushed outside of the window.

I noticed the bug in Moose while trying to import a model in a file, and was working on my laptop without any external screen (unsure whether or not it would happen on a bigger screen). I can however confirm that even if you maximize that dialog window, going deeply enough down the file tree will still end up pushing the buttons out of the screen.

It feels like a layout bug, so it might be better to fix it in `StFileDialogPresenter` so that this fix applies to all subclasses. However, I haven't investigated whether or not this bug could be present more generally across dialog windows, so this could also be something to look at.

Here is a screenshot of the full window after the buttons disappear:

Image

RomainDeg:
Fixes #1336

### DIFF
--- a/src/NewTools-FileBrowser/StFileSystemPresenter.class.st
+++ b/src/NewTools-FileBrowser/StFileSystemPresenter.class.st
@@ -110,15 +110,14 @@ StFileSystemPresenter >> connectPresenters [
 { #category : 'layout' }
 StFileSystemPresenter >> defaultLayout [
 
-	^ SpPanedLayout newLeftToRight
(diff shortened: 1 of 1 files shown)


guillep:
Merci @RomainDeg

PR #1338, opened 2026-01-31
Playground: Add delete page feature

jecisc:
This change adds a remove page icon. When we share the play-cache between multiple images it's common to have pages we do not want to keep. Instead of going to the folder to remove the files I propose to have an icon to do it from the playground directly

image

@estebanlm What do you think?

### DIFF
--- a/src/NewTools-Playground/StPlaygroundPage.class.st
+++ b/src/NewTools-Playground/StPlaygroundPage.class.st
@@ -119,6 +119,14 @@ StPlaygroundPage >> defaultObjectInspectorClass [
 	^ StPlaygroundPagePresenter
 ]
 
+{ #category : 'submorphs - add/remove' }
+StPlaygroundPage >> delete [
+	"Delete the corresponding page"
+
+	self fileReference ensureDelete.
+	(self fileReference withExtension: 'ombu') ensureDelete
+]
+
 { #category : 'accessing' }
 StPlaygroundPage >> ensureContentsFlushed [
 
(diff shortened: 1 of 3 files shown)


estebanlm:
lol, in fact I implemented that in the dev branch (which is being merged right now :P) 
(but, as a context menu option, which feels less intrusive). 
I would restrain this PR until the dev is merged and then evaluate if an icon helps (for me, is too intrusive, but I am not against per-se... just wait a bit and we'll evaluate it next week)

estebanlm:
also implementation has flaws: you should not use a button with a width (and in general, each time you add a width or a height, it is usually not really recommended: is there because there are  cases were you need it, but they are really few., 

instead, you add a button with expand: false and you set style 'small" to the button.
also, style 'flat" will make it look better.

jecisc:
Nice to know! I used `expand: false` at the beginning but it was take 3 times the space that it is taking with the #with:  I'll try later with styles.

For me, I'll want the button because I'll do this action a lot :P (Or I hesitate to modify the code to save only if the playground has more than one line of code. Because I type often just one line and it's rare I want to save those). But I can do that myself with a script. I don't need the default playground to have the button :P 

I would like other things for pages but that will be for the future, and I don't know if we want that in the default playground:
- Be able to give a title instead of using the first line
- Have categories (like I'd like to be able to say "This page is for Famix call graph" and be able to filter)
- Have a button to merge pages that have the same code (last version of one is the first version of the other)

But as I said, I might never have the time ahah

PR #1344, opened 2026-03-02
Cavrois fix for UI process order

AlexisCnockaert:
I did an investigation and the `applyWorldResize` was looking at existing windows without waiting the UI process execution, meaning it was looking for iceberg windows (if present in a profile) despite Iceberg not being initialized. It could also break other thing, having race conditions or any other thing, so I fixed it using the UIManager defer: .

Fixes[ #19149](https://github.com/issues/assigned?issue=pharo-project%7Cpharo%7C19149)

### DIFF
--- a/src/NewTools-Window-Profiles/CavroisWindowManager.class.st
+++ b/src/NewTools-Window-Profiles/CavroisWindowManager.class.st
@@ -728,36 +728,28 @@ CavroisWindowManager >> initializeBlocks [
 			self cellSize ifNotNil: [ :size |
 					self resetCells.
 					self createPlaceHolderGrid: size ].
-			self applyWorldResize ].
+			UIManager default defer: [ self applyWorldResize ] ].
 
 	WorldMorph dragActionBlock: [ :aClient |
-			((aClient isKindOf: SystemWindow) or:
-				 (aClient isKindOf: SpWindowMorph)) ifTrue: [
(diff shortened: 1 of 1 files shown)


src/NewTools-Window-Profiles/CavroisWindowManager.class.st
```diff
@@ -728,36 +728,28 @@ CavroisWindowManager >> initializeBlocks [
...
 							self cellDisplay.
 							dragProcess := [
 								               [ aClient owner notNil ] whileTrue: [
-										               self
-											               handleWindowDrag: aClient
-											               at: aClient position.
-										               (Delay forMilliseconds: 2) wait ] ] forkAt:
-								               Processor userBackgroundPriority.
+										               self handleWindowDrag: aClient at: aClient position.
+										               (Delay forMilliseconds: 2) wait ] ] forkAt: Processor userBackgroundPriority.
```
Ducasse, line 741: can you ask pablo or guille because I'm nervous to have  more processes around. 
The defer: is good because it is in the UI cycle.

AlexisCnockaert: Ok I will ask someone

AlexisCnockaert:
@Ducasse the processes were already integrated, it was for the drag n drop preview mode in the cavrois profiles

PR #1353, opened 2026-03-09
New scope tests based on introduced matching package scope

AlexisCnockaert:
These tests validate 3 behaviours: filtering based on prefix, `=`, and `hash` which are reimplemented on this new package. (The PR for this new scope is https://github.com/pharo-project/pharo/pull/19394)


AlexisCnockaert:
The broken tests on scopes are not related to this PR, even if it looks like :)
Other builds have the same broken tests in Pas

PR #1361, opened 2026-03-27
Fix method browser refreshing block for class refs

### LINKED ISSUE #19477: [New Method Browser]
- browse the sender of a method defined in a class and let it open
- then define a method with a reference to such a class.

AlexisCnockaert:
`hasLiteral:` looks for `GlobalVariable` and the method was given a symbol as argument. 
Fixes https://github.com/pharo-project/pharo/issues/19477


jecisc:
It would be nice to add a regression test if possible

AlexisCnockaert:
It is here :)

PR #1362, opened 2026-03-27
Adding-test-for-inspecting-outer-variable-in-clean-block

tesonep:
If we inpect or evaluate any expression using a variable in the outer context of a clean block it crashes the VM.

The fix in Pharo is PR #19487
This test will crash until that PR is integrated


tesonep:
Depends on https://github.com/pharo-project/pharo/pull/19487

Ducasse:
Guille did a large pass on all the logic and we got a better implementation of lookupVar: and friends

Ducasse:
I want to see the result of the integration tests with the new Pharo14.

PR #1371, opened 2026-04-03
[ Method Browser ] Add support for slot references in class

AlexisCnockaert:
Fully implemented 'Browser references of variable in class' for new Method Browser.
Also created a test to ensure good behavior


Ducasse:
Looks nice :)
Tx for this enhancements.

PR #1373, opened 2026-04-07
[ Method Browser ] Test result announcement refresh

AlexisCnockaert:
Now the method browser listen to Test announcements, meaning if we browse some test method and any of them is run elsewhere, the icons in the method browser will refresh based on test result. 
Also cleaned some code !


AlexisCnockaert:
Also a good improvement that could be done in the future is to only refresh the methods that were impacted by the announcement.
Because in case we have a method browser with a ton of methods it could be slow and bring problems

jecisc (approved):
This seems nice!

PR #1374, opened 2026-04-08
settings browser fixes

estebanlm:


estebanlm:
ok, now runs and the remaining error is not related

PR #1380, opened 2026-04-13
[Method Browser] New test based on previous PR

AlexisCnockaert:
I added 2 tests: - one is to ensure the title of the browser is updated well based on the filter
- the other one is to ensure method deletion is well handled in the methods collection by looking for the matching item

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -362,6 +362,12 @@ StMethodBrowser >> defaultLayout [
 	^ self perform: self class usingLayout
 ]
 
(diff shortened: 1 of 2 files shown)


Ducasse:
What would be good is to check that the index of a removed method garantees that when we remove a method we do not jump to the first one. 
Ideally we should be able to have 

```
A 
X B 
C
```
when X measn selected

and get 

```
A
XC
```

when we remove B


Now what if we remove the last one :) ok we select the previous one

And it would be great to have tests for this.

AlexisCnockaert:
This is an improvement I've made in the previous PR ! i will add a test for this

AlexisCnockaert:
I put everything in #1381

PR #1381, opened 2026-04-14
[Method Browser] Improve responsiveness to system announcements/new tests

AlexisCnockaert:
I added mulitple tests, as well of announcement handle fix for the `MethodBrowser`:
-Fix handleMethodAdded:: Added a check to prevent duplicate entries if a method is already present in the browser list.

Test Suite Enhancements:
-Added tests for MethodAdded (ensuring correct filtering/adding).
-Added tests for MethodModified (ensuring removal of stale references).
-Added tests for MethodRemoved (ensuring list consistency and selection updates).
-Added a UI test for the window title filtering ratio.


Ducasse:
Looks good!

PR #1383, opened 2026-04-15
[Method Browser] Proposing to save changes for dirty methods

### LINKED ISSUE #19553: Closing method browser with edited code doesn't propose to save changes
When modifying code in a method browser text editor, then closing it either with the cross, or cmd+w, it discards the code automatically. It would be good to propose saving the changes before closing.

AlexisCnockaert:
This code will work when all the `SpCodePresenter` stuff will be able to handle it, for now it don't understand when a method is dirty.  I just prepared the ground. 
However as Esteban said: "My intention is to create a meta-code-presenter (SpPharoEditorPresenter) that will handle all higher level stuff that can be modeled composing presenters", I will not propose a quick solution here, his idea is way better. 
Since he has a prototype, I think we can wait.

Fixes [#19553](https://github.com/pharo-project/pharo/issues/19553)

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -559,6 +559,10 @@ StMethodBrowser >> initializePresenters [
 StMethodBrowser >> initializeWindow: aWindowPresenter [
 
 	super initializeWindow: aWindowPresenter.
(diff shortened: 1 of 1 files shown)


Ducasse:
Tx for the notice

AlexisCnockaert:
Now it works since the editor has been implemented !

PR #1384, opened 2026-04-15
[Method Browser] Cleaning code and adding button feature

AlexisCnockaert:
- Removed unused method for technical debt
- Some methods used `System Navigation new` instead of the default one, fixed
- New `Run Tests` button in the toolbar to run all test methods present in the browser instance
- Removed `windowIsClosing`, there was 2 different call sites to handle window closing in the code, I moved all the logic in `initializeWindow:`


Ducasse:
argh we get a conflict.

AlexisCnockaert:
I solved it :)

PR #1391, opened 2026-04-20
use the code editor presenter on StMethodBrowser

estebanlm:


AlexisCnockaert:
Hello @estebanlm, just a few things that I reviewed:
I intentionally updated `handleMethodModified:`  to handle modifications of any method in the list (not only the selected one) using `detect:ifNone:`. Your version reverts to checking only selectedMethod, which reintroduces the bug fixed in #1379 where modifying a non-selected sender from another browser was not reflected in the list.
`handleMethodAdded:` was also updated to preserve selection via `defer: `due to a GTK timing issue I think(https://github.com/pharo-project/pharo/issues/19545) where selectIndex: was called before the widget finished rebuilding. Your version removes this fix.
Also, the `includes` for handling duplicates in added methods has been removed, and no methods are being added at all now based on your changes I believe.

Could you keep both the existing fixes and your changes please ?

estebanlm:
I can try :P

estebanlm:
I merged with your changes, @AlexisCnockaert ... but I am not sure that thing is working as we would expect, as your changes do not replace the element in the same position, and then we get lost :)
Anyway, can you check the changeset in case I missed something else?

AlexisCnockaert:
I will fix the replacing position thing when this will be merged I think its better.
I checked the change set and its good! it only misses both previous `defer:` on the index selection.

And very last thing it missed the merge with `handleMethodAdded:`, I think a good mix for our versions would be
```
handleMethodAdded: anAnnouncement
    | item selectedMethod editedText |
    self isDisplayed ifFalse: [ ^ self ].
    refreshingBlock ifNil: [ ^ self ].
    item := anAnnouncement method.
    (self shouldRefreshItem: item fromAnnouncement: anAnnouncement) ifFalse: [ ^ self ].
    (item methodClass isNotNil and: [ item methodClass isObsolete not ]) ifFalse: [ ^ self ].

    editedText := textPresenter isDirty ifTrue: [ textPresenter text ].
    selectedMethod := self selectedMethod.

    (self methods includes: item) ifFalse: [
        self methods: (self methods add: item asFullRingDefinition; yourself) ].

    self defer: [ self selectIndex: (self methods indexOf: selectedMethod ) ].
    self selectedMessage: selectedMethod .

    editedText ifNotNil: [
        textPresenter text: editedText.
        textPresenter markDirty ]
  ```

AlexisCnockaert (approved):

estebanlm:
I have no idea why the debugger tests are failing :/
(but is not related to this change)

PR #1393, opened 2026-04-20
[Method Browser] Refactoring command migration

AlexisCnockaert:
Start of refactoring commands migration from Calypso to Spec !
This is a good setup in order to use them for the MethodBrowser ( not for playground or anything else yet, we need to provide a unified API so the commands don't raise errors).
If there are broken tests it may be because the command classes need to be merged in Spec repository.


AlexisCnockaert:
Yes all failing tests will pass when [Spec PR](https://github.com/pharo-spec/Spec/pull/1892) will be merged

PR #1402, opened 2026-04-23
New command for Spec refactorings

### LINKED ISSUE #1295: The comments of all the class side methods of StMessageBrowser should be revised
They all say the same

```
browse: aCollection
	"Special Version that sets the correct refreshing Block for Implentors Browser"
```

And well this is not true.

AlexisCnockaert:
Moved all refactoring commands from Spec to New Tools since they inherit from `StCommand`

I also added new commands:
- Push Up/Down variable command
- Remove variable command
- Protect variable command
- Move IV to class side

Also Fixes #1295 by fixing bad method comments and adding ones


jecisc:
Are the errors related to the fact that a Spec PR needs to be merged first? Multiple of them are failing

AlexisCnockaert:
Yes it is due to the spec PR

PR #1405, opened 2026-04-24
[Method Browser] Baseline update and fixes

AlexisCnockaert:
- New spec command: `Rename Variable` + test!
- Refactoring commands are now hidden when they cannot be executed.
- Fixed CI tests.
- Added the refactoring command packages by default in the baseline. 
- Updated initialization so the code editor can be marked dirty. 
- Lastly, precised the title when there is no results (i.e `Implementors of #printOn: No Results`).

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodCodePresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodCodePresenter.class.st
@@ -50,6 +50,7 @@ StMethodCodePresenter >> refactoringCommandsGroup [
 		StPushUpVariableCommand.
 		StPushDownVariableCommand.
 		StRemoveVariableCommand.
-		StProtectVariableCommand } do: [ :cmdClass | group register: (cmdClass forSpecContext: methodBrowser) ].
+		StProtectVariableCommand.
+		StRenameVariableCommand } do: [ :cmdClass | group register: (cmdClass forSpecContext: methodBrowser) ].
 	^ group
 ]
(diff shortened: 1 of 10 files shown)


jecisc:
The navigation system tests are not related?

AlexisCnockaert:
> The navigation system tests are not related?

No it is linked to recent changes on click events I believe

src/NewTools-MethodBrowsers/StMethodCodePresenter.class.st
```diff
@@ -50,6 +50,7 @@ StMethodCodePresenter >> refactoringCommandsGroup [
 		StPushUpVariableCommand.
 		StPushDownVariableCommand.
 		StRemoveVariableCommand.
-		StProtectVariableCommand } do: [ :cmdClass | group register: (cmdClass forSpecContext: methodBrowser) ].
+		StProtectVariableCommand.
+		StRenameVariableCommand } do: [ :cmdClass | group register: (cmdClass forSpecContext: methodBrowser) ].
 	^ group
```
jecisc, line 55: This can be change in a future PR but I don't like hardcoded list like this. Ideally we should collect the commands. Because if I want to customize my IDE with my own command, I cannot we a hard coded list :(

AlexisCnockaert: i understand, I was thinking of changing it too

PR #1406, opened 2026-04-27
[Refactoring command migration] Extract method

AlexisCnockaert:
Implemented `Extract Method` command for Spec !


Ducasse:
Youpi!

Ducasse:
unrelated broken tests

PR #1415, opened 2026-04-30
[Method Browser] Use `debug` instead of `run` for tests

### LINKED ISSUE #1412: New method browser should use debug: instead of run: to execute the tests
or we should have a way to choose
may be icon is run and the menu offers run and debug

AlexisCnockaert:
- It allows for the user to directly interact with failing/breaking tests instead of browsing them by himself.
- Also updated the remove method command in the method browser, now it calls the refactoring ( with the conditions)

Fixes #1412

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
@@ -247,10 +247,7 @@ StMethodListPresenter >> executeTestFor: aMethod [
 
 	testClasses do: [ :each | suite addTest: (each selector: selector) ].
 
-	result := TestResult new.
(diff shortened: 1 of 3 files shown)


Ducasse:
I would prefer that we improve the testResult so that the debug: implements the right stuff.

AlexisCnockaert:
@Ducasse I improved the debug in https://github.com/pharo-project/pharo/pull/19628, so this can work !

jecisc:
I think this can cause a problem. The updated method seems to be used by the "Run tests" button. If we execute all tests we might not want to have a debut

AlexisCnockaert:
@jecisc I agree with it, I updated the changes for it :)

PR #1417, opened 2026-05-05
[Refactoring Spec Command] Class side registry for commands

AlexisCnockaert:
- Instead of hardcoded commands in an array, commands can now register themselves via a registry. It allows a user to configurate its own commands in an easier way
- These commands are also sorted by priority 
- Precised existing spec command name (R/T to know if it is refactoring/transformation


AlexisCnockaert:
Broken test for Rename method commands is related to https://github.com/pharo-project/pharo/pull/19620 (need to be merged)

jecisc (approved):

PR #1418, opened 2026-05-05
New opening messages for Trait/Pool users

AlexisCnockaert:
- Created 2 opening methods looking for  users of a trait and of a shared pool
- As well increased the default extent of the browser, since it can contain very large names

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -195,6 +195,35 @@ StMethodBrowser class >> browseSendersOfAll: aCollectionOfSymbols [
 ]
 
 { #category : 'opening' }
+StMethodBrowser class >> browseUsersOfSharedPool: aPoolClass [
+
+	| methods classes poolVarNames |
+	classes := aPoolClass poolUsers.
+	poolVarNames := aPoolClass classVarNames.
+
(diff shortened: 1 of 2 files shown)


src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -194,6 +194,35 @@ StMethodBrowser class >> browseSendersOfAll: aCollectionOfSymbols [
...
 ]
 
+{ #category : 'opening' }
+StMethodBrowser class >> browseUsersOfSharedPool: aPoolClass [
+
+	| methods classes poolVarNames |
+	classes := aPoolClass poolUsers.
+	poolVarNames := aPoolClass classVarNames.
+
+	methods := classes flatCollect: [ :cls |
```
jecisc, line 204: Ideally we should avoid abbreviations and have #class instead of #cls and #method instead of #m (in the next method). 

It makes the reading of code easier

But I'll merge anyway. It's more a general comment. Abbreviations make us lost time when reading code

jecisc (approved):

src/NewTools-MethodBrowsers/StMethodBrowser.class.st
AlexisCnockaert: I understand, I will keep that in mind for the future

PR #1420, opened 2026-05-06
[Method browser] Add deprecation display for method

### LINKED ISSUE #19531: [Method browser] implementors should show deprecated methods
- when the method browser shows deprecated method we should get a visual feedback.

AlexisCnockaert:
- Deprecated methods are now marked like they were in Calypso
- Added a regression test!
- Fixes https://github.com/pharo-project/pharo/issues/19531

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
@@ -290,17 +290,10 @@ StMethodListPresenter >> initializePresenters [
 		addColumn: (SpColumnViewColumn new
 				 title: 'Selector';
 				 sortFunction: #selector ascending;
-				 bind: [ :p :item |
-						 | label |
-						 label := self selectorOf: item.
-
-						 item isDeprecated
-							 ifTrue: [
-									 p label: (label asText
-												  addAttribute: TextEmphasis struckOut;
-												  addAttribute: (TextColor color: Color gray);
-												  yourself) ]
-							 ifFalse: [ p label: label ] ];
+				 bind: [ :labelPresenter :item |
+						 labelPresenter label: (self selectorOf: item).
+						 labelPresenter displayStrikethrough: [ item isDeprecated ].
+						 labelPresenter displayColor: [ item isDeprecated ifTrue: [ Color gray ] ] ];
 				 yourself);
 		addColumn: (SpColumnViewColumn new
 				 title: 'Package';
--- a/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
@@ -290,7 +290,17 @@ StMethodListPresenter >> initializePresenters [
 		addColumn: (SpColumnViewColumn new
 				 title: 'Selector';
 				 sortFunction: #selector ascending;
-				 bind: [ :p :item | p label: (self selectorOf: item) ];
+				 bind: [ :p :item |
+						 | label |
+						 label := self selectorOf: item.
+
+						 item isDeprecated
+							 ifTrue: [
+									 p label: (label asText
+												  addAttribute: TextEmphasis struckOut;
+												  addAttribute: (TextColor color: Color gray);
+												  yourself) ]
+							 ifFalse: [ p label: label ] ];
 				 yourself);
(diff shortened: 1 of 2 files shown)


src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
```diff
@@ -290,7 +290,17 @@ StMethodListPresenter >> initializePresenters [
...
+						 | label |
+						 label := self selectorOf: item.
+
+						 item isDeprecated
+							 ifTrue: [
+									 p label: (label asText
+												  addAttribute: TextEmphasis struckOut;
+												  addAttribute: (TextColor color: Color gray);
+												  yourself) ]
+							 ifFalse: [ p label: label ] ];
```
jecisc: I am not sure this will work with all backends of Spec.

Is there a reason not to add #displayStrikethrough: on Spec as we discussed?

AlexisCnockaert: The thing is SpColumnViewColumn  don't support the display messages, it work only with SpringTableColumn.
And we work with a ColumnView here for the tests buttons, we cannot switch

jecisc: I would like the opinion of @estebanlm on this one :) 

A little context for esteban: the goal of Alexis here is to strikethrouh the deprecated methods of the StMessageBrowser. 

Is it already possible with Spec's API? If not, does this proposition work?

estebanlm: somehow, striketrough is missing in the spec api.
I am doing a PR now because indeed, directly dealing with the Text element is not the way to go.

estebanlm: voilà: https://github.com/pharo-spec/Spec/pull/1906

AlexisCnockaert: Okay but its still the same problem: "The thing is SpColumnViewColumn don't support the display messages, it work only with SpringTableColumn or other presenters.
And we work with a ColumnView here for the tests buttons, we cannot switch"

estebanlm: no, because now you have a label presenter (that you call "p", and by the way, I do not recommend calling like that your variables. There I would call it `aPresenter` or `labelPresenter` to be more clear of what it is or what we expect there), and a label presenter uses `SpTDecoratedText`... which means you can do something like this:
```smalltalk
labelPresenter displayStrikethrough: [  item isDeprecated ]   
```

EDIT: and you change the color of a label by adding/removing styles.

AlexisCnockaert: Ok i did not know that, thanks Esteban

AlexisCnockaert:
@estebanlm could you check if my changes are okay now please ?

src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
```diff
@@ -290,7 +290,10 @@ StMethodListPresenter >> initializePresenters [
 		addColumn: (SpColumnViewColumn new
 				 title: 'Selector';
 				 sortFunction: #selector ascending;
-				 bind: [ :p :item | p label: (self selectorOf: item) ];
+				 bind: [ :labelPresenter :item |
+						 labelPresenter label: (self selectorOf: item).
+						 labelPresenter displayStrikethrough: [ item isDeprecated ].
+						 labelPresenter displayColor: [ item isDeprecated ifTrue: [ Color gray ] ] ];
```
jecisc: #displayColor: will do nothing if the result of the block is nil?

estebanlm: probably. Also this is bad since the color then is hardcoded and it will remain the same if you change the style, and that's obviously not good. 
The solution is to add/remove an style

AlexisCnockaert: Okay I checked how styles work and it is cleaner.
But I saw that in the old browser the deprecated methods didn't have any color, so I propose to just mark the method name with the strike through if you all agree with it.
If so, we can merge this !

PR #1422, opened 2026-05-07
[Method Browser] Allow multiple selections

### LINKED ISSUE #19647: StMethodBrowser misses multiple selection
The previous message browser allowed multiple selection but the new one misses this feature. 

This is useful in some cases like removing methods, browsing senders or implementors of a group of methods, etc.

@AlexisCnockaert

AlexisCnockaert:
- Adjusted the command calls so they handle a collection if there is multiple selection
- The browser just had a single selection before that is why
- Fixes https://github.com/pharo-project/pharo/issues/19647

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -765,17 +765,17 @@ StMethodBrowser >> navigateInPlace [
 StMethodBrowser >> navigateTo: aType value: anObject [
 
 	self navigateInPlace ifFalse: [
-			aType == #implementors ifTrue: [ ^ self class browseImplementorsOf: anObject ].
-			aType == #implementorsAll ifTrue: [ ^ self class browseImplementorsOfAll: anObject ].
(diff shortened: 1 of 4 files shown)


Ducasse:
StMethodBrowserTest
 ✗ #testHandleMethodRemovedRemovesMatchingMethod (129ms)
 ✗ #testHandleMethodRemovedUpdatesSelection (141ms)

ReRenameClassCommandTest
 ✗ #testRenameClassCommandSuccess (665ms)

AlexisCnockaert:
@Ducasse tests are fixed

jecisc:
This seems OK but I see that we have some ifs checking if we have 1 or more elements.
Can't we have only one API always getting a collection? And we pass a collection of 1 element if we selected only one?

Is there a change of behavior between executing the code with 1 element or a collection with 1 element?

AlexisCnockaert:
You are right, we can pass a collection each time and call the same thing

AlexisCnockaert:
@jecisc I fixed this

jecisc (approved):

PR #1423, opened 2026-05-07
Refactoring command initialization strategy

AlexisCnockaert:
- Moved the command registry to the command package
- Lazy init for the commands in the getter instead of `initialize`


jecisc (approved):

jecisc:
I am not yet sure which failing tests are related or not since I'm coming back from holidays. Can you check the failing tests of Pharo-Integration to confirm they are not linked?

AlexisCnockaert:
The same tests are broken in an other PR, all unrelated.
Also it took 8/9min to run the CI, I wonder what happened

PR #1436, opened 2026-05-11
feat: added resume button to StProcessBrowser

terryc321:
A use case below of a digit clock morph can suspend and resume from playground but not process browser

morph := StringMorph new.
morph openInWorld. 
process := [ [ true ] whileTrue: [ 
    morph contents: Time now print24.
    1 second wait 
] ] forkAt: Processor systemBackgroundPriority .
process suspend.
process resume.
process terminate.
morph delete.

### DIFF
--- a/src/NewTools-ProcessBrowser/StProcessBrowser.class.st
+++ b/src/NewTools-ProcessBrowser/StProcessBrowser.class.st
@@ -217,7 +217,7 @@ StProcessBrowser >> buildProcessToolbarActions [
 		yourself
 ]
 
-{ #category : 'testing' }
+{ #category : 'actions' }
 StProcessBrowser >> canResumeSelectedProcess [
 
 	| process |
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks for your PR. We will check it.

src/NewTools-ProcessBrowser/StProcessBrowser.class.st
```diff
@@ -445,6 +460,18 @@ StProcessBrowser >> refreshProcessList [
 	processListPresenter updateItemsKeepingSelection: self model processList
 ]
 
+{ #category : 'as yet unclassified' }
```
Ducasse: Can you recategorise this method?

terryc321: moved two methods for resume process to actions , not sure if this is correct.

PR #1440, opened 2026-05-12
doc:tidy StPlayground comment in microdown format

terryc321:
exchange !!! with ### and *class:X* with `X` . 
perhaps a tool can automate this , maybe have a viewer before after , what it looks like in microdown


Ducasse:
Tx!

PR #1441, opened 2026-05-13
[Method Browser] Senders of a recursive method should not appear

### LINKED ISSUE #19688: StMethodBrowser should show no senders for recursive methods that only are call by themselves
If we have a recursive method that is only called by itself (because is recursive). The StMethodBrowser should say that there are no senders (because they are not).

To reproduce:
Press cmd + n to get the method browser on the senders of a recursive method

Image

AlexisCnockaert:
- Having a recursive method only called by itself should show 'No Results'
- Fixes https://github.com/pharo-project/pharo/issues/19688

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -311,8 +311,12 @@ StMethodBrowser class >> registerToolsOn: registry [
 StMethodBrowser class >> sendersOf: aSymbol [
 	"Do not use aSymbol senders because it is a disguised global"
 
-	^ SystemNavigation default allSendersOf: aSymbol
-	
+	| senders nonRecursive |
+	senders := SystemNavigation default allSendersOf: aSymbol.
+
+	nonRecursive := senders reject: [ :method | method selector = aSymbol ].
+
+	^ nonRecursive
 ]
 
 { #category : 'icons' }


jordanmontt:
So fast! Thanks!

However, this will not work because you are checking only if the **selector** is the same. This will break Polymorphic calls. For example, if we have this method, which is not a recursion.

Capture d’écran 2026-05-13 à 14 20 24

It will show that there are no senders. 

Capture d’écran 2026-05-13 à 14 21 21


We should not only check the selector but also that the receiver is self. It would be nice to have tests, too

jecisc:
@AlexisCnockaert Hi Alexis,

We discussed with Sebastian and we have a few comments.

First, as sebastian said above, we should not consider only the selector but also the origin class. I know we do not have it yet, but we will need to provide it anyway in order to make the scope works. So maybe you can fix both issues at the same time?

Second, we agree with Sebastian that the browser should say "no result" if there is only one result that is recursive. But if it has other results, we should keep the recursive calls in the list. Because sometimes you can chain some queries about senders and implementors and we do not want to lose some results. 

Third, I'd like to have tests on this so that we know this is intentional :) 

In the meantime I think we should revert this PR because I'm not sure the fix will be "easy" since we need the origin class?

jordanmontt:
I agree with Cyril!

- For the first point, I think that is enough that we just check is the receiver is `self`.  For that we will probably need to check the AST.
- For the second point, I also agree that if the method has more senders than the recursive call, we show **all** of them including the recursive call. If not, we show empty
- Also, yes, since we heavily depend on this feature I will say that we need tests.
- I agree also wit the reverting of this PR

AlexisCnockaert:
Ok I understand, well I need to find a way to pass the origin class to the browser so I can fix both issues, thanks for explaining all points.

jordanmontt:
@AlexisCnockaert we can work on this on the Pharo Sprint

AlexisCnockaert:
@jordanmontt ok I'm down :)

PR #1442, opened 2026-05-13
Create a call back for the CLI debugger

NathanMalenge:
PR depending on https://github.com/pharo-project/pharo/pull/19619
While working on a CLI debugger, I had to refactor debugger-Model beacause it used UIManager directly.
So the goal of this PR is to avoid this dependency, using callBacks. We have to have a callBack class in newTools because it's what calls debugger-Model. 
 We use StApplication and spec to open a dialog instead of UIManager


src/NewTools-Debugger/GraphicDebugDelegate.class.st
```diff
@@ -0,0 +1,26 @@
...
+GraphicDebugDelegate >> chooseTraitOverwriteFor: aClass trait: aTrait [
+
+	^ UIManager default
+		  chooseFrom: {
+				  ('Create copy in ' , aClass name).
+				  ('Compile Trait method in ' , aTrait name) }
+		  values: {
+				  aClass.
+				  aTrait }
+		  title: 'Where do you want to compile this trait method?'
```
jecisc: We want to remove the usage of UIManager so I'm not sure this code is a good idea

UI interactions should be managed by the UI. Ideally we should open a spec presenter/dialog via the StPharoApplication. But for that we need to get it from the UI

guillep: This was before hardcoded in the model (DebugSession, DebugContext) that was "UI agnostic" and did not know the presenter/app.

This "delegate" introduces a callback now. The model will call a callback, have a default value when it's not set, and the UI will set a specific callback now. Allowing us to cut the dependency.

Now, for this case, the delegate will have a reference to the presenter and be able to access the application from there. Now, we should see how this method is replaced in the app

src/NewTools-Debugger-Tests/TStDebuggerExtensionTestClass.class.st
```diff
@@ -8,6 +8,12 @@ Class {
 	#tag : 'Model'
 }
 
+{ #category : 'as yet unclassified' }
+TStDebuggerExtensionTestClass class >> acceptsPredicate: aStDebuggerContextPredicate [
+
+	^ true
+]
+
```
guillep: This is strange, did you touch this?

guillep (approved):
Nice! would you mind adding a PR description? It should at least say
 - that this depends on the other PR you did in Pharo
 - what it is supposed to do and why

src/NewTools-Debugger-Tests/TStDebuggerExtensionTestClass.class.st
NathanMalenge: i did not, or not in purpose

PR #1447, opened 2026-05-18
Bring History and Copy commands to method browser

### LINKED ISSUE #19691: StMethodBrowser is missing the iceberg history and Copy selected method names command
In the previous method browser we could use commands to open the iceberg history browser (that browse the commit history VS the current image history that e have with "versions") and the command to copy the selected method names. The second command is useful to send some methods names we discussing with people.

AlexisCnockaert:
- Added previous browser commands in the new method browser: `Commit history` and `Copy to clipboard`
- Also added the Flip icon to the toolbar
- Fixes https://github.com/pharo-project/pharo/issues/19691

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowserToolbarPresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowserToolbarPresenter.class.st
@@ -66,6 +66,7 @@ StMethodBrowserToolbarPresenter >> initializePresenters [
 
 	toolbarPresenter add: (self newToolbarButton
 			 label: 'Flip';
+			 icon: (self iconNamed: #glamorousRefresh);
 			 action: [ self doFlipLayout ])
 ]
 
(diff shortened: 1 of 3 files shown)


src/NewTools-MethodBrowsers/StMethodBrowserToolbarPresenter.class.st
```diff
@@ -66,6 +66,7 @@ StMethodBrowserToolbarPresenter >> initializePresenters [
 
 	toolbarPresenter add: (self newToolbarButton
 			 label: 'Flip';
+			 icon: (self iconNamed: #glamorousRefresh);
```
jecisc:
```suggestion
			 icon: (self iconNamed: #refreshCircling);
```

We want to avoid icons with Glamorous in the name

jecisc (approved):
This will be good once the suggestion applied

PR #1448, opened 2026-05-18
[Method Browser] Introduce the caller method to the browser

AlexisCnockaert:
- Each instance of the browser now knows its caller, which will be useful for managing scopes and other stuff
- The caller is initialized from each existing constructor
- Added a test to check if `callerMethod` is set well!

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -18,7 +18,8 @@ Class {
 		'refreshingBlock',
 		'textPresenter',
 		'highlight',
-		'filterPresenter'
+		'filterPresenter',
+		'callerMethod'
 	],
 	#classVars : [
 		'NavigateInPlace',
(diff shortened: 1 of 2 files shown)


jecisc:
If possible, for this one I'd like the opinion of @MarcusDenker 

He knows better than me the effects of using the context

src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -274,6 +276,20 @@ StMethodBrowser class >> navigateInPlaceSettingsOn: aBuilder [
 		parent: #tools
 ]
 
+{ #category : 'instance creation' }
+StMethodBrowser class >> newFromContext [
```
Ducasse, line 280: To me this is strange to use thisContext here. We do not need the call stack.
The tool could just store the currently displayed information before spawning another tool.

MarcusDenker: yes, using reflection like this is not good. This can be useful in case one does a minimal intrusive change e.g. for some experiment, but here we can do any refactoring needed to avoid using reflection.

AlexisCnockaert: The thing is we want to know from which place the browser is opened, the displayed information (I.E selected method) can be anything.
We discussed with Cyril that we should know the caller to initialize the scopes correctly as the old browser did.

AlexisCnockaert: > yes, using reflection like this is not good. This can be useful in case one does a minimal intrusive change e.g. for some experiment, but here we can do any refactoring needed to avoid using reflection.

So what could be the alternative here, passing the callerMethod by adding it to every constructor ?

PR #1449, opened 2026-05-18
Hierarchy building fix

### LINKED ISSUE #19681: StMethodBrowser does not build the hierarchy right when we filter or apply a scope
If we apply a filter or a scope the list does not build the hierarchy right. 

For example:

Image

With a scope

Or with a filter:

Image

AlexisCnockaert:
- The hiearchy was not built right because `methods:` cancelled the sort at initialization
- Applied a sort block to sort items by hierarchy indentation
- Deleted unused methods and `sortFunction:` for the classes, since we want the hierarchy built from the start. We can still sort by method or package
- Test to validate hierarchical sorting
- Fixes https://github.com/pharo-project/pharo/issues/19681

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
@@ -359,7 +359,7 @@ StMethodListPresenter >> methods: aCollection [
 
 	self cacheHierarchyForClasses: aCollection.
 	listPresenter items:
-		(cachedHierarchy keys asSortedCollection: [ :a :b | self should: a havePriorityOver: b ]) asOrderedCollection.
+		(cachedHierarchy keys asSortedCollection: [ :method :otherMethod | self should: method havePriorityOver: otherMethod ]) asOrderedCollection.
 	allMessages := listPresenter items.
 
 	listPresenter listSize > 0 ifTrue: [ self defer: [ self selectIndex: 1 ] ]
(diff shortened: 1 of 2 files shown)


src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
```diff
@@ -295,6 +257,22 @@ StMethodListPresenter >> handleTestResultChanged: anAnnouncement [
...
+{ #category : 'sorting' }
+StMethodListPresenter >> hierarchySortBlock [
+
+	^ [ :a :b |
+		  | hierarchyA hierarchyB minSize answer |
+		  hierarchyA := cachedHierarchy at: a ifAbsent: [ #(  ) ].
+		  hierarchyB := cachedHierarchy at: b ifAbsent: [ #(  ) ].
+
+		  minSize := (hierarchyA size min: hierarchyB size) - 1.
+		  answer := nil.
```
jecisc: Answer is already nil

src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
```diff
@@ -397,10 +374,10 @@ StMethodListPresenter >> methods [
 StMethodListPresenter >> methods: aCollection [
 
 	self cacheHierarchyForClasses: aCollection.
-	listPresenter items: cachedHierarchy keys asOrderedCollection.
+	listPresenter items: (cachedHierarchy keys asSortedCollection: self hierarchySortBlock ) asOrderedCollection.
```
jecisc: I would prefer something like:

`cachedHierarchy keys asSortedCollection: [ :method :otherMethod | self should: method havePriorityOver: otherMethod ]`

Having a method returning a block is not nice

AlexisCnockaert: I get it, I commited your suggestion

PR #1450, opened 2026-05-18
add build info to system reporter.

### LINKED ISSUE #19653: Getting Pharo version in P14 is not really logical and a mess
**1st Problem:**   
The system reporter in Pharo 13 when clicking on "Image" shows 
- Pharo version:  "Pharo13.1.0SNAPSHOT" 
- Image build version: Pharo-13.1.0+SNAPSHOT.build.732.sha.e84a2d15c798f5f7da482f14dcc7fb488b9aedbc (64 Bit)

Image

so one can reproduce which Pharo version is used and where the image exactly came from.

In Pharo 14 the build info is gone and the same Pharo version is shown two times:

Image

so in Pharo 14 (as it is now) one is unfortunately not anymore able to reproduce the exact image build directly from the system reporter. This is clearly a bug.

**2nd Problem:**   
Running 

```Smalltalk
Smalltalk vm version
```

return the VM version as expected - this is fine.

Running 

```Smalltalk
Smalltalk image version
```
returns "Pharo14.0.0SNAPSHOT" but this #version call has now been deprecated (see the depreaction in SmalltalkImage>>#version which was introduced in P14)

So why do we deprecate (and later remove) to ask for the image version while asking for the vm version is still valid?

This is not logical, not consistent and contrary to the goal to easily ask for the image and vm version in the same easy way.
To me it makes no sense.

estebanlm:
fixes https://github.com/pharo-project/pharo/issues/19653

### DIFF
--- a/src/NewTools-SystemReporter/StSystemReporterPresenter.class.st
+++ b/src/NewTools-SystemReporter/StSystemReporterPresenter.class.st
@@ -71,14 +71,14 @@ StSystemReporterPresenter >> fillTopics [
 
 { #category : 'utils' }
 StSystemReporterPresenter >> isLinux [
-	
-	^Smalltalk os isLinux
+
+	^ OSPlatform current isLinux
 ]
 
 { #category : 'utils' }
 StSystemReporterPresenter >> isWindows [
 
-	^Smalltalk os isWindows
+	^ OSPlatform current isWindows
 ]
 
 { #category : 'building report' }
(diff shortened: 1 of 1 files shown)


estebanlm:
commit also includes the fix of a couple of deprecations.

Ducasse:
good. I fixed many of them but not all.

Ducasse:
Unrelated broken test

PR #1455, opened 2026-05-21
[Method Browser] No browser opening when no results

### LINKED ISSUE #19663: StMethodBrowser when there is no result
In the previous message browser, if we had 0 result the browser did not open but an inform was raised to let the user know there was no result.

With the StMethodBrowser we are opening the browser and display that there is no result.

Personally, I prefer the old way to not have to close a useless browser. 

Do other have an opinion on this?

AlexisCnockaert:
- Instead of opening a new empty browser when no results are found for a query, the user gets informed that no there are no results. This avoid an extra window to pop up
- Fixes https://github.com/pharo-project/pharo/issues/19663

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -839,6 +839,13 @@ StMethodBrowser >> numberOfElements [
 	^ methodList ifNil: [ 0 ] ifNotNil: [ methodList numberOfElements ]
 ]
 
+{ #category : 'showing' }
+StMethodBrowser >> open [
+
+	filterPresenter unfilteredItems isEmpty ifTrue: [ ^ self application inform: 'No result found' ].
+	^ super open
+]
+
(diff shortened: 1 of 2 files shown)


src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -916,6 +916,13 @@ StMethodBrowser >> numberOfElements [
 	^ methodList ifNil: [ 0 ] ifNotNil: [ methodList numberOfElements ]
 ]
 
+{ #category : 'showing' }
+StMethodBrowser >> open [
+
+	filterPresenter unfilteredItems isEmpty ifTrue: [ ^ self application inform: 'No result found' ].
```
jecisc:
```suggestion
	allMethods ifEmpty: [ ^ self application inform: 'No result found' ].
```

jecisc (changes requested):

jecisc (approved):

PR #1460, opened 2026-05-24
[Method Browser] Fix `locationOf:` unknown red square

### LINKED ISSUE #19706: StMethodBrowser leads to red square of death
Image

### LINKED ISSUE #19722: Red square of death while using the StMethodBrowser
I was refactoring methods in a StMethodBrowser and I got a red square of death. 

I have no idea what code change produced this bug, but I have a screenshot of the stack trace. Maybe we can add a guard somewhere to prevent it.

Image

AlexisCnockaert:
- Since we can't identify when the red square of death happens, I added a fix to prevent a key not found error
- Fixes https://github.com/pharo-project/pharo/issues/19722
- Fixes https://github.com/pharo-project/pharo/issues/19706

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodListPresenter.class.st
@@ -361,14 +361,16 @@ StMethodListPresenter >> listPresenter [
 { #category : 'accessing' }
 StMethodListPresenter >> locationOf: anItem [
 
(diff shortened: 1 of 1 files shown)


jecisc (approved):
Good since we do not know the scenario causing the bug.

I'm not sure about the method name because reading the name I was expecting an index as a return value. But it's not related to this PR

PR #1465, opened 2026-05-29
commit two "Smalltalk os" refactorings

MarcusDenker:
just some removals of deprecated usage of "Smalltalk os"


MarcusDenker:
as these are just deprecation rewrites, I merge

PR #1470, opened 2026-06-02
[Method Browser] No results when only recursive call

### LINKED ISSUE #19688: StMethodBrowser should show no senders for recursive methods that only are call by themselves

AlexisCnockaert:
- As discussed with Cyril and Sebastian, we want to have no result for a recursive call when the origin class is the caller, and that there is only one result
- If they are multiple results, we include this recursive call in the list
- Fixes https://github.com/pharo-project/pharo/issues/19688


jecisc (commented):
I think it would be nice to have a comment to explain the code in the method

src/NewTools-MethodBrowsers/StMethodBrowserTest.class.st
```diff
@@ -353,6 +353,48 @@ StMethodBrowserTest >> testReuseWindowCheckboxInitialState [
...
 
+{ #category : 'tests' }
+StMethodBrowserTest >> testSendersHidesRecursiveCallWhenOnlySender [
+	"If the only sender of a method is itself, no result should be shown"
+
+	| browser senders |
+	[
+		self class compile: 'tmpRecursiveMethod ^ self tmpRecursiveMethod' classified: 'tests-protocol'.
+
+		senders := (StMethodBrowser sendersOf: #tmpRecursiveMethod) reject: [ :m | m methodClass = self class ].
```
jordanmontt: we shouldn't just test that `(StMethodBrowser sendersOf: #tmpRecursiveMethod)` gives an empty list? Why are you doing the filtering youself ?

src/NewTools-MethodBrowsers/StMethodBrowserTest.class.st
```diff
@@ -353,6 +353,48 @@ StMethodBrowserTest >> testReuseWindowCheckboxInitialState [
...
+{ #category : 'tests' }
+StMethodBrowserTest >> testSendersKeepsRecursiveCallWhenOtherSendersExist [
+	"If a method has other senders besides itself, keep all including the recursive call"
+
+	| browser senders |
+	[
+		self class compile: 'tmpRecursiveMethod2 ^ self tmpRecursiveMethod2' classified: 'tests-protocol'.
+		self class compile: 'tmpOtherCaller ^ self tmpRecursiveMethod2' classified: 'tests-protocol'.
+
+		senders := (StMethodBrowser sendersOf: #tmpRecursiveMethod2) reject: [ :m |
```
jordanmontt: same here

src/NewTools-MethodBrowsers/StMethodBrowserTest.class.st
AlexisCnockaert: @jordanmontt it is because the test method is sending `tmpRecursiveMethod` itself, so in this test there are two senders, `tmpRecursiveMethod` and the test method. I filtered the senders so it could test the right thing.

AlexisCnockaert: Tell me if you have an other solution, but I figured this out since I had the problem.

src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -885,11 +885,24 @@ StMethodBrowser >> methods: compiledMethods asReferencesToVariable: aVariable in
...
+	 If callerClass is set and all senders are recursive calls (i.e. the method only calls itself),
+	 the list is left empty so the browser displays 'No result found'.
+	 If other senders exist alongside the recursive call, all senders are kept in the list
+	 to avoid losing results when chaining sender/implementor queries."
+
+	| filteredMethods |
+	filteredMethods := self callerClass ifNil: [ compiledMethods ] ifNotNil: [ :caller |
+			                   | withoutSelf |
+			                   withoutSelf := compiledMethods reject: [ :method |
+					                                  method methodClass instanceSide = caller instanceSide and: [
```
jordanmontt: why you do `method methodClass instanceSide = caller instanceSide` instead of `method methodClass = caller` ?

src/NewTools-MethodBrowsers/StMethodBrowserTest.class.st
```diff
@@ -353,6 +353,49 @@ StMethodBrowserTest >> testReuseWindowCheckboxInitialState [
...
+		self assert: browser allMethods isEmpty ] ensure: [ self class removeSelector: #tmpRecursiveMethod ]
+]
+
+{ #category : 'tests' }
+StMethodBrowserTest >> testSendersKeepsRecursiveCallWhenOtherSendersExist [
+	"If a method has other senders besides itself, keep all including the recursive call"
+
+	| browser senders |
+	[
+		self class compile: 'tmpRecursiveMethod2 ^ self tmpRecursiveMethod2' classified: 'tests-protocol'.
```
jordanmontt: instead of compiling methods and then removing the selector, don't you think is better if we have a "mock" class to use as the test fixture? So you put this code `tmpRecursiveMethod2 ^ self tmpRecursiveMethod2` inside a mock class inside this package. Like that you don't need to compile and remove the methods

jordanmontt:
@AlexisCnockaert 
> @jordanmontt it is because the test method is sending tmpRecursiveMethod itself, so in this test there are two senders, tmpRecursiveMethod and the test method. I filtered the senders so it could test the right thing.

something like we could do, it is also a hack, is to concatenate the selector. For example `#tmpRecursive , #Method`. Like that that selector will not have no senders and we do not need to do the filtering. It is also a hack but something else does not occurs me

AlexisCnockaert:
> @AlexisCnockaert
> 
> > @jordanmontt it is because the test method is sending tmpRecursiveMethod itself, so in this test there are two senders, tmpRecursiveMethod and the test method. I filtered the senders so it could test the right thing.
> 
> something like we could do, it is also a hack, is to concatenate the selector. For example `#tmpRecursive , #Method`. Like that that selector will not have no senders and we do not need to do the filtering. It is also a hack but something else does not occurs me

I never tried it and it works well thanks!

jordanmontt:
Looks good for me! There are just two more things (sorry haha!)

- I will change the name `StMethodBrowserData` to something that states that this is just a dumb class with methods that we only use for testing. The current name gives the impression that this is some kind of data holder class.
- The method `browser methods: compiledMethod asSendersOfAll: selectors.` expects a collection of selectors. In the current tests, we are only testing the behavior for **one** selector. I think that it is important to add some test that also tests this feature (recursive calls being left out) when we send a collection with more than one selector

src/NewTools-MethodBrowsers/StMethodBrowserDummyClass.class.st
```diff
@@ -0,0 +1,39 @@
+Class {
+	#name : 'StMethodBrowserDummyClass',
+	#superclass : 'Object',
+	#traits : 'TData',
+	#classTraits : 'TData classTrait',
```
jordanmontt, line 5: Why do we need to use a trait in here?

AlexisCnockaert: It is related to an other test, if we browse the senders of a trait method we should only show the trait method itself and reject the users of the method

jordanmontt (approved):

jordanmontt:
Looks good to me. We could think of other features in another PR.

What do you think? @jecisc

PR #1471, opened 2026-06-03
Fix: browsing senders from traits showed users too

AlexisCnockaert:
- When browsing senders of a method coming from a trait, the methods coming from users of the trait were shown too
- Now it is not !
- New test for method exclusion (trait)

### DIFF
--- a/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
+++ b/src/NewTools-MethodBrowsers/StMethodBrowser.class.st
@@ -338,11 +338,14 @@ StMethodBrowser class >> registerToolsOn: registry [
 
 { #category : 'private' }
 StMethodBrowser class >> sendersOf: aSymbol [
-	"Do not use aSymbol senders because it is a disguised global"
+	"Do not use aSymbol senders because it is a disguised global.
+	When browsing senders of a trait method we only want to have the method trait itself.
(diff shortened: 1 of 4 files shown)


src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -340,8 +340,10 @@ StMethodBrowser class >> registerToolsOn: registry [
 StMethodBrowser class >> sendersOf: aSymbol [
 	"Do not use aSymbol senders because it is a disguised global"
 
-	^ SystemNavigation default allSendersOf: aSymbol
+	| allSenders |
+	allSenders := SystemNavigation default allSendersOf: aSymbol.
 	
+	^ allSenders reject: [ :sender | sender isFromTrait and: [ sender origin ~= sender methodClass ] ]
```
jecisc, line 349: Do we need both checks? Aren't they redundant?

AlexisCnockaert: We need both because if we just put `isFromTrait`, we wouldn't have any methods coming from a Trait.
And the second check allows to know whether the method is coming from a trait OR used from a trait.

jecisc: Can you add this as a comment please? :)

PR #1473, opened 2026-06-08
Fixing senders of

Ducasse:


src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -906,7 +906,7 @@ StMethodBrowser >> navigateTo: aType value: anObject [
 			aType == #implementorsAll ifTrue: [
 				self class methods: (anObject flatCollect: [ :sel | self class implementorsOf: sel ]) asImplementorsOfAll: anObject ].
 			aType == #sendersAll ifTrue: [
-				self class methods: (anObject flatCollect: [ :sel | self class sendersOf: sel ]) asSendersOfAll: anObject ].
+				self class new methods: (anObject flatCollect: [ :sel | self class sendersOf: sel ]) asSendersOfAll: anObject ].
```
AlexisCnockaert:
```suggestion
				self class browse: (anObject flatCollect: [ :sel | self class sendersOf: sel ]) asSendersOfAll: anObject from: nil ].
```

src/NewTools-MethodBrowsers/StMethodBrowser.class.st
```diff
@@ -906,7 +906,7 @@ StMethodBrowser >> navigateTo: aType value: anObject [
 			aType == #implementorsAll ifTrue: [
 				self class methods: (anObject flatCollect: [ :sel | self class implementorsOf: sel ]) asImplementorsOfAll: anObject ].
```
AlexisCnockaert:
```suggestion
				self class browse: (anObject flatCollect: [ :sel | self class implementorsOf: sel ]) asImplementorsOfAll: anObject from: nil ].
```

AlexisCnockaert (commented):
It should use the class constructors since we want to open a new one

src/NewTools-MethodBrowsers/StMethodBrowser.class.st
jecisc: Should we declare a "from" here? Does it makes sense?

Ducasse:
Ok for me :)

src/NewTools-MethodBrowsers/StMethodBrowser.class.st
AlexisCnockaert: The thing is I removed the constructors that did not have `from` for multiple implementors/senders.
Either we recreate them, we leave it like that, or we call `from:nil` if that makes sense.
What do you think ?

jecisc: I don't have a strong opinion on this. The simplest is to pass nil I guess

AlexisCnockaert (changes requested):

AlexisCnockaert:
ok @Ducasse you just have to apply changes!

PR #1478, opened 2026-06-11
Hide Refactoring commands in browser

AlexisCnockaert:
- When can't be run, a command is hidden instead of being disabled. This makes the list cleaner
- Also introduced `Move method to class side` for Spec + test for the command !


AlexisCnockaert:
Tests are failing for commands because the ui-tests package is not loaded in latest image (baseline restructuration)

Ducasse:
Alexis I discussed with esteban and it would be good that you sycn with him next week.

AlexisCnockaert:
We discussed a solution for shortcuts in the method browser and this PR is not related

PR #1481, opened 2026-06-14
Update rewriter tools

Gabriel-Darbord:
Some improvements to the RewriterTools package:

1. `StRewriterScopeSelectorPresenter` opened debugger when executing [`openDialog`](https://github.com/pharo-spec/NewTools/blob/003ab41d42689e139d8d3af8e7e726a5c240ec4c/src/NewTools-RewriterTools/StRewriterScopeSelectorPresenter.class.st#L139) because its class does not implement `windowExtent`. Simply removing the method makes it behave as intended.

2. `StRewriterRuleEditorPresenter` produced unnamed windows because its title method was incorrectly named `windowTile` instead of `windowTitle`. I also renamed the window to match the name that appears in the "Library" World menu.

3. The help presenter now uses rich Text, which seems to align with the original intention. I also added section headers, code blocks, and simplified the prose a bit where I thought it would help.


Ducasse:
Thanks!!!! This is great to see this help!

PR #1486, opened 2026-06-25
Fix search of StBrowserSearchPresenter

jecisc:
The search was always using #name but it is possible to define another dislpay block. Now it reuses the display block to do the search. 

Before if we did a find on a method list, the search was also matching the class name. Now it does it only on the selector


Gabriel-Darbord (approved):
LGTM.
I was wondering if the test should use `StBrowserSearchPresenter` directly instead of a test double, but the behavior is consistent regardless of what methods we add/remove in the future, **as long as we don't rename the class**.

Gabriel-Darbord (approved):
More future-proof, thanks!

PR #1491, opened 2026-06-26
Use test as debug callback and set to recompile new method

### LINKED ISSUE #19819: testRecompileMethodToInBlockContextWithDeadHome
- We check and the failure is not related to cleanblock
We changed 

```
methodWithDeadBlockContext
	| x |
	^ [ x := 2. (1 + x) printString ]
```

and the test fails too.

The point is that `wantsToBroweAndRecompileMethodNotFoundInStack:` as used here

```
blockContextNotFoundForMethod: aMethod withSourceCode: sourceCode inContext: aContext

	| result newMethod |
	self halt.
	result := self debugCallBack wantsToBroweAndRecompileMethodNotFoundInStack: aMethod. 
	
	"possible return values are true | false | nil"
	result ifNil: [ ^ self ].
	result ifFalse: [ ^ self ].
	
	newMethod := aContext recompileCurrentMethodTo: sourceCode notifying: nil inSession: self.
	newMethod ifNotNil: [ aContext browseMethod: newMethod ]
```

 is defined as 

```
DebugSession >> wantsToBroweAndRecompileMethodNotFoundInStack: message
	"call back method to avoid dependencies in debugger-model"
	"by default it returns false the modified code is not recompiled"
	
	^ false
```

guillep:
With the introduction of the recompile callback, when a callback is not set the default value does a different thing than the UI (which was tested).

Fixes https://github.com/pharo-project/pharo/issues/19819#issuecomment-4808373903

### DIFF
--- a/src/NewTools-Debugger-Tests/StDebuggerClientModelTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerClientModelTest.class.st
@@ -195,10 +195,11 @@ StDebuggerClientModelTest >> setResult [
 StDebuggerClientModelTest >> setSessionAndDebuggerModelForMethod: aMethod inContext: aContext [
 
 	| dummyActionModel |
-	session ifNotNil: [ session clear ].
+	"session ifNotNil: [ session clear ]."
 	dummyActionModel := (StTestDebuggerProvider new debuggerWithContext:
 		                     aContext) debuggerClientModel.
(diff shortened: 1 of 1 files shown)


Ducasse:
@guillep don't you think that it would be good that the debugSession can control the wantsRecompile or not?
Because hardcoding false looks strange.

PR #1492, opened 2026-06-26
fix tests and typos in selector

Ducasse:
Should be integrated once https://github.com/pharo-project/pharo/pull/19832 is integrated


Ducasse:
So I close it because it was fixed in a different way
@guillep

PR #1499, opened 2026-07-11
Fixing prof stef p14

Ducasse:


Ducasse:
I should see how I can backport it to P15

capsulecorplab:
Why were all the periods removed in https://github.com/pharo-spec/NewTools/pull/1499/changes/afa6effd3dfb6885c96db1c18b8893abe81166ea ?

Ducasse:
Because periods in Pharo are separators and not terminators (you do not have to a period at the end of an expression).
However when I did that (removing the useless periods) I broke the pretty print of the code because the page of ProfStef are stored "comment" expr "comment" expres and the expres was not pretty printed well. 
We should get an explicit notebook like structure to delimit clearly the expressions and in such case the terminal period will be unnecessary. So after removing the periods I reintroduced them. 
Am'i clear?

S

capsulecorplab:
Ah, that makes sense. And yes, I see the periods were reintroduced back in https://github.com/pharo-spec/NewTools/pull/1501

PR #1500, opened 2026-07-11
Improve the profiler

jecisc:
The code of the profiler is currently a huge mess. Here is a first cleaning. 

I propose to integrate this changes in Pharo 14 because it fixes layouts problems where buttons where outside of the presenter. So we should forward it to P15 also. 
On top of this, I'll try to fix more things in Pharo 15 on my free time.

- Remove hardcoded sizes in layouts to fix hidden actions
- Do not display the code input if we open the profiler by profiling a piece of code directly
- Move connections between presenters to #connectPresenters
- Remove some subclasses of widgets
- Use #instantiate: to pass the application to sub presenters
- Make presenters inherit from StPresenter instead of SpPresenter
-  Make the default width of the presenter large enough to contain all buttons displayed by default
- Move layout from #initializePresenters to #defaultLayout


jecisc:
@estebanlm Do you want to review or should I merge?
This bring a lot of improvement to the profiler to make it kinda usable for P14 :)

estebanlm:
it looks ok. It made me remember I need to fix the number input (is broken in the max/min stuff). But is not related with this :)

PR #1505, opened 2026-07-23
Adding

Ducasse:
package: 'NewTools-Spotter-Extensions' with: [ spec requires: #( 'NewTools-Spotter' 'NewTools-SettingsBrowser') ];


Ducasse:
This is strange that the build is not kicked.

PR #1083, opened 2025-04-02
[WIP] Message Browser

tesonep:
Making the Message browser in a good shape

- Requires https://github.com/pharo-project/pharo/pull/18082
- Then it allows us to merge https://github.com/pharo-project/pharo/pull/18085
- There are some issues open that should be nice before integrating this or at least before integrating the Pharo PR:
   - #1086 
   - #1087 
   - #1088 
   - #1090
   - #1097
   - #1100

PR #1507, opened 2026-08-03
Harvested tests for the method browser

AlexisCnockaert:
Took tests from https://github.com/pharo-spec/NewTools/pull/1083 and adapted them to the current API so the tests for the method browser are solidified
- Multiple tests : scope switching, addition/removal of method event

### DIFF
--- a/.github/workflows/newtools-all.yml
+++ b/.github/workflows/newtools-all.yml
@@ -6,9 +6,9 @@ env:
 
 on:
   push:
-    branches: [ pharo-15 ]
+    branches: [ Pharo15 ]
   pull_request:
-    branches: [ pharo-15 ]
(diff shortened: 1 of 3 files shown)


jecisc:
In order for the CI to run, could you update the CI scripts to run on pharo-15 branch and testing on Pharo15 instead of pharo-14 branch and P14 please? You can do it in this PR

AlexisCnockaert:
I need to modify both newtools-... yml files ?

jecisc:
Yes! one to run only new tools test and one to run most Pharo tests :)

Ducasse:
nice ;)

jecisc:
Apparently SmalltalkCI is not yet working

PR #1511, opened 2026-08-06
Forward Port fixes done in P14 to P15

guillep:
Magic cherrypicking stuff


Ducasse:
So I will integrate

Ducasse:
I wonder if I did not go too fast. 
Right now opening the setting browser I get a DNU settingTreeRoots

PR #1521, opened 2026-09-07
Cleaning as string

Ducasse:
just one less user ot asString (it blocks my Pharo 15 PR)


Ducasse:
asString should not get confused with printString or displayString.

Ducasse:
Broken tests are unrelated

PR #1522, opened 2026-09-08
use default (and correct) layout

estebanlm:
fixes https://github.com/pharo-project/pharo/issues/19973


Ducasse:
Tx!
I read it to learn how do you address it.

S

estebanlm:
in this case, the problem was the hardcoded height that was breaking the layouting.
while this should not happen (and this is arguably a bug), hardcoding values should not be used.
I also took the oportunity to "modernise" the layout (it was obviously implemented by migrating directly old spec definitions (from when they were on class side) to new style.

Ducasse:
Thanks for the explanation. I reread the PR twice to try to understand and I could not find where it was hardcoded so I could not have fixed it.

PR #1531, opened 2026-09-18
fix shadow variable method (and fix incorrect use of delete)

estebanlm:


estebanlm:
errors now are not related

